PR #10, opened 2018-02-16
Parallelize IPP with rayon

cathieyun:
Each commit is a different configuration of the calculations for `G, H`.

On my own laptop, results for `make_ipp_64` are:
7c811db: unoptimized: `17,958,887 ns/iter (+/- 2,645,491)`
1654069: x-axis parallelization: `10,427,160 ns/iter (+/- 321,816)` (updated)
655ab83: y-axis parallelization: `17,827,652 ns/iter (+/- 2,091,841)`
e230f13: x-axis and y-axis parallelization: `10,461,115 ns/iter (+/- 711,309)` (updated)

x-axis = parallelizing the calculation of each index of the vectors `G` and `H`.
y-axis = parallelizing the calculation of `G` and `H`, with a `join()`.

Next steps:
- get zoomzoom github access so we can run the benchmarks there
- pick one of the 4 (or potentially more) options to use


hdevalence:
I think this branch was accidentally overwritten; closing this PR since it's stale

PR #12, opened 2018-03-01
Aggregated range proofs

cathieyun:
done:
- [x] get basic state machine prover & verifier to work
- [x] restructure states to have separate commitment parts (suggestions from @oleganza )
- [x] renaming suggestions from @oleganza 
- [x] make `Proof::prove` shortcut for single player version
- [x] generate `n*m` generators for `G`, `H`
- [x] update proof generation to index into generators correctly
- [x] use random oracle API instead of `commit` function
- [x] commit to other parts of the proof (`n`, `m` , `V`, etc)
- [x] combine the state machine range proof approach with the aggregated proof math
- [x] write an aggregator / "manager" (e.g. for getting commitments, returning challenges?)
- [x] update verifier to work for aggregated proofs
- [x] change j indexing to start at 0 instead of 1
- [x] always pass in RNG (never create)
- [x] store input in statement (instead of just inputCommitment)

further work (in a later PR - #57 )
- [x] where to fit errors in the states (if one party's share doesn't parse correctly, or if we have to check whether things were valid)
- [x] can we do per-party share validity checks?
- [ ] proof & commitments passed into verifier separately (?)

NOTE:
since we are integrating the IPP, and the IPP requires vectors of length `2^k` where `k` is an integer, we have to choose the number of parties we use accordingly.


src/multi_range_proof.rs
```diff
@@ -0,0 +1,728 @@
...
+        let y = self.transcript.challenge_scalar();
+        let z = self.transcript.challenge_scalar();
+
+        (
+            DealerAwaitingPoly {
+                transcript: self.transcript,
+                n: self.n,
+            },
+            y,
+            z,
```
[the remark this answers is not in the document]
cathieyun: 👍

src/multi_range_proof.rs
```diff
@@ -0,0 +1,728 @@
...
+    /// Blinding factor for the synthetic commitment to \\(t(x)\\)
+    pub t_x_blinding: Scalar,
+    /// Blinding factor for the synthetic commitment to the inner-product arguments
+    pub e_blinding: Scalar,
+    /// Proof data for the inner-product argument.
+    pub ipp_proof: inner_product_proof::Proof,
+
+    // FIXME: don't need if doing inner product proof
+    pub l: Vec<Scalar>,
+    pub r: Vec<Scalar>,
```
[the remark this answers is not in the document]
cathieyun: 👍

src/multi_range_proof.rs
```diff
@@ -0,0 +1,728 @@
...
+            value_commitments,
+            A,
+            S,
+            T_1,
+            T_2,
+            t_x: t,
+            t_x_blinding,
+            e_blinding,
+            ipp_proof,
+            l: l_vec,
```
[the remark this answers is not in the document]
cathieyun: Oh i see now :)

src/multi_range_proof.rs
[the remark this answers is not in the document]
cathieyun: 👍

oleganza:
(moving discussion from Slack)

You could use an enum for the various errors that you return, so it's easy to embed, say, an index of a faulty party. 

```
#[derive(Debug)]
enum Error {
    RangeIsNotPowerOfTwo, // n is not a power of 2
    PartiesCountIsNotPowerOfTwo, // m is not a power of 2
    InconsistentProofShare(usize), // party j:size provided a vector of length != n
    InvalidProofShare(usize), // party j:usize provided an invalid proof share
    ...
}
```

oleganza:
(moving discussion from Slack)

To make aggregated codebase friendly to a single-party use case (e.g. i'm creating a transaction with several outputs all by myself), we could delay verification of individual proof shares by making the dealer return `.receive_shares() -> (Proof, AggregatedProofVerifier)`. Then, one can verify the proof using a usual method (`proof.verify`) or via `<AggregatedProofVerifier>.verify()` which remembers all proof shares and checks individual shares only if the complete proof is not correct.

cathieyun:
@hdevalence PTAL - I fixed up the basic aggregated proof, and moved the error checking to another PR (#57 )

src/aggregated_range_proof/dealer.rs
```diff
@@ -0,0 +1,174 @@
...
+/// Dealer is an entry-point API for setting up a dealer
+pub struct Dealer {}
+
+impl Dealer {
+    /// Creates a new dealer with the given parties and a number of bits
+    pub fn new(
+        transcript: &mut ProofTranscript,
+        n: usize,
+        m: usize,
+    ) -> Result<DealerAwaitingValues, &'static str> {
```
hdevalence: Should this return a `Result` if it's always going to be `Ok`?

src/aggregated_range_proof/dealer.rs
```diff
@@ -0,0 +1,174 @@
...
+}
+
+impl DealerAwaitingValues {
+    /// Combines commitments and computes challenge variables.
+    pub fn receive_value_commitments(
+        self,
+        transcript: &mut ProofTranscript,
+        vc: &Vec<ValueCommitment>,
+    ) -> (DealerAwaitingPoly, ValueChallenge) {
+        // TODO: test that vc is length `m`.
```
hdevalence: I think for now it would be OK to just `assert!` this (causing a panic if it fails) for now and deal with the error handling in #57

src/aggregated_range_proof/dealer.rs
```diff
@@ -0,0 +1,174 @@
...
+        let mut A = RistrettoPoint::identity();
+        let mut S = RistrettoPoint::identity();
+
+        for commitment in vc.iter() {
+            // Commit each V individually
+            transcript.commit(commitment.V.compress().as_bytes());
+
+            // Commit sums of As and Ss.
+            A += commitment.A;
+            S += commitment.S;
```
hdevalence: Why do we commit only sums of `A`s and `S`s? Is it because only the sums are available to the verifier?

Is it possible for colluding parties to form `A`s and `S`s that would cancel each other in the sum?

hdevalence: I think the answer is maybe "no", because in order for the aggregated proof to verify, the `A`s and `S`s have to be formed with respect to each party's generators, and those generators are supposed to be orthogonal. So colluding parties who formed `A`s and `S`s that were related to each other would have to know relations between the supposedly orthogonal generators.

src/aggregated_range_proof/dealer.rs
cathieyun: I was setting up the function to potentially return an error - but then I moved the error handling to PR #57 so this isn't needed here.

src/aggregated_range_proof/dealer.rs
cathieyun: yes, it's because only the sums are available to the verifier. (we can commit to individual V's on the other hand because they are available to the verifier).

And yes, your guess/explanation makes sense

src/aggregated_range_proof/tests.rs
```diff
@@ -0,0 +1,106 @@
...
+            &generators.all(),
+            value_challenge.y,
+        )
+    }
+
+    fn test_u32(m: usize) {
+        let mut rng = OsRng::new().unwrap();
+        let mut transcript = ProofTranscript::new(b"AggregatedRangeProofTest");
+
+        let v: Vec<u64> = iter::repeat(())
```
oleganza: why not `(0..m).into_iter().map(|_| rng.next_u32() )`?

src/aggregated_range_proof/tests.rs
```diff
@@ -0,0 +1,106 @@
...
+            .collect();
+        let rp = create_multi(&mut rng, v, 32);
+        assert!(rp.verify(&mut rng, &mut transcript).is_ok());
+    }
+
+    fn test_u64(m: usize) {
+        let mut rng = OsRng::new().unwrap();
+        let mut transcript = ProofTranscript::new(b"AggregatedRangeProofTest");
+
+        let v: Vec<u64> = iter::repeat(()).map(|()| rng.next_u64()).take(m).collect();
```
oleganza: Same as above: `repeat().map().taken(m)` feels a bit backwards. Although, it's purely a matter of style, so I'd defer to more experienced rustaceans.

src/aggregated_range_proof/dealer.rs
```diff
@@ -0,0 +1,174 @@
...
+        m: usize,
+    ) -> Result<DealerAwaitingValues, &'static str> {
+        transcript.commit_u64(n as u64);
+        transcript.commit_u64(m as u64);
+        Ok(DealerAwaitingValues { n, m })
+    }
+}
+
+/// When the dealer is initialized, it only knows the size of the set.
+pub struct DealerAwaitingValues {
```
hdevalence: Minor nit, but maybe we should name the `AwaitingX` states so that they match the inputs of their processing step, like `DealerAwaitingValueCommitments` which takes `ValueCommitment`s, etc.

src/aggregated_range_proof/dealer.rs
```diff
@@ -0,0 +1,174 @@
...
+        gen: &GeneratorsView,
+        y: Scalar,
+    ) -> Proof {
+        let value_commitments = proof_shares
+            .iter()
+            .map(|ps| ps.value_commitment.V.clone())
+            .collect();
+        let A = proof_shares
+            .iter()
+            .fold(RistrettoPoint::identity(), |A, ps| {
```
hdevalence: this seems like a reason to implement `Sum` and `Product` upstream in `curve25519-dalek`

src/aggregated_range_proof/dealer.rs
```diff
@@ -0,0 +1,174 @@
...
+        transcript.commit(T1.compress().as_bytes());
+        transcript.commit(T2.compress().as_bytes());
+
+        let x = transcript.challenge_scalar();
+
+        (DealerAwaitingShares { n: self.n }, PolyChallenge { x })
+    }
+}
+
+pub struct DealerAwaitingShares {
```
hdevalence: nit: could be `DealerAwaitingProofShares` for similar reasons as above

src/aggregated_range_proof/party.rs
```diff
@@ -0,0 +1,222 @@
...
+        // Return next state and all commitments
+        let value_commitment = ValueCommitment { V: self.V, A, S };
+        let next_state = PartyAwaitingValueChallenge {
+            n: self.n,
+            v: self.v,
+            v_blinding: self.v_blinding,
+
+            j,
+            generators: self.generators,
+            value_commitment: value_commitment.clone(),
```
hdevalence: The explicit clone can be removed if `ValueCommitment` implements `Copy`

hdevalence:
:+1: from me too. This is great. The only comments I have are minor cosmetic changes. I think there are also some warnings that could be cleaned up.

@cathieyun if you're working on writing notes/docs I can patch up the warnings and cosmetic stuff, or we could do that as a later PR, whichever you want to do. WDYT?

cathieyun:
@hdevalence that'd be great if you could patch up the warnings & do the renamings & merge in the PR when that is done! I'll keep working on the docs.

hdevalence:
The cosmetic fixes are pushed above. @cathieyun maybe you can just double check that I didn't nuke anything, and then merge/squash (whichever you want)?

cathieyun:
merged!

Issue #22, opened 2018-04-03
Batch verification for the rangeproofs


oleganza:
Given N rangeproofs it is possible to verify them in a single multi-scalar multiplication by merging the relevant iterators.

hdevalence:
Since we moved to aggregated proofs, I think #27 is stale and has to be rewritten to handle the aggregated-proof verification checks.

hdevalence:
I think we should aim for the same thing that we did with aggregation: internally, verification should be implemented as `k`-batch verification of `m`-aggregated proofs of `n`-bits each, and we provide specializations of `k, m = 1`.

hdevalence:
A general conceptual comment that we should think about:

With single proofs, we have a conceptually nice story about how the proof transcripts work: the transcript implements the Fiat-Shamir transform for the interactive protocol between the prover and the verifier. Because we pass a `&mut ProofTranscript` into all of the functions, proofs and verifications are composable. 

For instance, the range proof code sends messages and receives challenges, then passes the same `&mut ProofTranscript` into the inner-product code, so that the IPP can be implemented without reference to the RP code, but still have all of its challenges bound to the parent protocol.

However, this batch verification implementation takes a single transcript, and then clones it to replay each proof's messages. This works for the case where all proofs are created with the same initial transcript state (e.g., a message with a domain separation label), but breaks composability.  Maybe that's OK, but in that case the verification function should take ownership of the `ProofTranscript` object, since it's no longer suitable for use after the function has been called.

On the other hand, if we try to preserve composability by passing a separate `&mut ProofTranscript` for each proof to be verified, then we can't properly handle failures: if one proof fails to verify, causing the whole batch to fail, then attempting to audit which proof was invalid requires rolling back each transcript state.

Since batch verification is a performance optimization that's only useful when processing large volumes of proofs, maybe we should wait to implement it until we have a better idea about how we want to use and compose proofs, so that we have a better understanding of the details of the above questions.

oleganza:
👍  thanks for the sketch.

PR #27, opened 2018-04-05
Batch verification for range proofs (replaced)

oleganza:
**REPLACED BY #86.**

This adds batch-verification to any number of range-proofs. The API ensures that all rangeproofs use the same width `n`.

Addresses #22.

Performance depends on Pippenger implementation (not merged yet): https://github.com/dalek-cryptography/curve25519-dalek/pull/129


oleganza:
This PR is replaced by a newer rewrite in #86.

PR #57, opened 2018-04-23
Aggregated range proof enhancements

cathieyun:
Things added:
- [x] state outputs are Results, to allow for error returns
- [x] individual proof share validity checking
- [x] ProofBlame struct for ^ and tests for ProofBlame
- [x] check that m and n are valid powers of 2 and that all dealer states take in `m` items

### DIFF
--- a/src/aggregated_range_proof/dealer.rs
+++ b/src/aggregated_range_proof/dealer.rs
@@ -18,6 +18,12 @@ impl Dealer {
         m: usize,
         transcript: &mut ProofTranscript,
     ) -> Result<DealerAwaitingValueCommitments, &'static str> {
+        if !n.is_power_of_two() {
+            return Err("n is not a power of two")
+        }
+        if !m.is_power_of_two() {
+            return Err("m is not a power of two")
+        }
         transcript.commit_u64(n as u64);
         transcript.commit_u64(m as u64);
         Ok(DealerAwaitingValueCommitments { n, m })
--- a/src/aggregated_range_proof/dealer.rs
+++ b/src/aggregated_range_proof/dealer.rs
@@ -18,11 +18,11 @@ impl Dealer {
         m: usize,
         transcript: &mut ProofTranscript,
     ) -> Result<DealerAwaitingValueCommitments, &'static str> {
-        if !n.is_power_of_two() {
-            return Err("n is not a power of two");
+        if !n.is_power_of_two() || n > 64 {
+            return Err("n is not valid: must be a power of 2, and less than or equal to 64");
         }
         if !m.is_power_of_two() {
-            return Err("m is not a power of two");
+            return Err("m is not valid: must be a power of 2");
         }
         transcript.commit_u64(n as u64);
         transcript.commit_u64(m as u64);
--- a/src/aggregated_range_proof/dealer.rs
+++ b/src/aggregated_range_proof/dealer.rs
@@ -61,13 +61,15 @@ impl DealerAwaitingValueCommitments {
 
         let y = transcript.challenge_scalar();
         let z = transcript.challenge_scalar();
+        let value_challenge = ValueChallenge { y, z };
 
         Ok((
             DealerAwaitingPolyCommitments {
                 n: self.n,
                 m: self.m,
+                value_challenge: value_challenge.clone(),
             },
-            ValueChallenge { y, z },
+            value_challenge,
         ))
     }
 }
@@ -76,6 +78,7 @@ impl DealerAwaitingValueCommitments {
 pub struct DealerAwaitingPolyCommitments {
     n: usize,
     m: usize,
+    value_challenge: ValueChallenge,
 }
 
 impl DealerAwaitingPolyCommitments {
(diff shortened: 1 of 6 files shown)


cathieyun:
(moving discussion with @oleganza from Slack)

You could use an enum for the various errors that you return, so it's easy to embed, say, an index of a faulty party.

```
#[derive(Debug)]
enum Error {
    RangeIsNotPowerOfTwo, // n is not a power of 2
    PartiesCountIsNotPowerOfTwo, // m is not a power of 2
    InconsistentProofShare(usize), // party j:size provided a vector of length != n
    InvalidProofShare(usize), // party j:usize provided an invalid proof share
    ...
}
```

cathieyun:
(moving discussion with @oleganza from Slack)

To make aggregated codebase friendly to a single-party use case (e.g. i'm creating a transaction with several outputs all by myself), we could delay verification of individual proof shares by making the dealer return .receive_shares() -> (Proof, AggregatedProofVerifier).

Then, one can verify the proof using a usual method (proof.verify) or via <AggregatedProofVerifier>.verify() which remembers all proof shares and checks individual shares only if the complete proof is not correct.

So the two possible flows become something like that:

Single party: compute proof, discard the verifier, use proof in the transaction and do no further checks.
```
let (proof, _) = dealer.receive_shares(...);
publish(proof)
```
Multiple parties: compute proof, verify proof, ban the faulty parties
```
let (proof, verifier) = dealer.receive_shares(...);
match verifier.verify(proof) {
   Ok(_) => publish(proof)
   Err(Error::InvalidProofShare(let j)) => ban_party(j)
}
```
That said, we can make API even more rigid if we have a different entry point for the single party. Then we can wrap all the dealer-party interaction under a single call and also assume that all parties are correct (because they all are encapsulated).

cathieyun:
moving comment from @hdevalence from PR #12:

We should also figure out how to handle the powers-of-2 restriction on the IPP protocol. My feeling so far is that we should require that the bitlengths of the ranges n to have power-of-two-sizes 8 / 16 / 32 /64, and m should be a power of two also. This means that the ranges that can be proved in aggregate are the same as the ranges that can be proved not-in-aggregate.

src/aggregated_range_proof/dealer.rs
```diff
@@ -0,0 +1,237 @@
...
+    m: usize,
+}
+
+impl DealerAwaitingValues {
+    /// Combines commitments and computes challenge variables.
+    pub fn receive_value_commitments(
+        self,
+        value_commitments: &Vec<ValueCommitment>,
+        transcript: &mut ProofTranscript,
+    ) -> Result<(DealerAwaitingPoly, ValueChallenge), (DealerAwaitingValues, &'static str)> {
```
hdevalence: The error type of this `Result` is a `(DealerAwaitingValues, &'static str)`. I guess this is so that if there's an error, the code orchestrating the dealer state can retry and keep going with the protocol.

Is it a good idea to try to resume the protocol from an intermediate dealer state, or to fail without the possibility of resumption (i.e., return just the error part `&'static str`, and consume the dealer state so that the protocol has to start over)?

I think failing without resumption would be simpler, and I'm not sure whether it would make sense to retry here, since the error is caused by not having enough value commitments (which seems like something the orchestrating code should be ensuring when it collects them from whereever).

cathieyun:
PTAL @hdevalence @oleganza

src/aggregated_range_proof/dealer.rs
```diff
@@ -81,23 +106,50 @@ impl DealerAwaitingPolyCommitments {
...
         gen: &GeneratorsView,
-        y: Scalar,
-    ) -> Proof {
+        transcript: &mut ProofTranscript,
+    ) -> Result<(Proof, Vec<ProofBlame>), &'static str> {
+        if self.m != proof_shares.len() {
+            return Err("Length of proof shares doesn't match expected length m");
+        }
+
+        let mut proof_blame = Vec::new();
```
oleganza: If it's a vector, it should probably be plural, and since it's a per-share checker, should it have "share" in its name? How about `share_verifiers: Vec<ProofShareVerifier>`?

cathieyun: that sounds much better! can do

src/aggregated_range_proof/messages.rs
```diff
@@ -168,23 +257,16 @@ impl Proof {
...
-    let sum_of_powers_of_z = util::exp_iter(*z)
-        .take(m)
-        .fold(Scalar::zero(), |acc, x| acc + x);
-
-    let zz = z * z;
+    (z - z * z) * sum_y - z * z * z * sum_2 * sum_z
+}
 
-    (z - zz) * sum_of_powers_of_y - z * zz * sum_of_powers_of_2 * sum_of_powers_of_z
+// XXX this could be more efficient, esp for powers of 2
```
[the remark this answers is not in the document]
oleganza: The intuition behind this is to split the polynomial `1+x+x^2+...+x^(n-1)` into two halves as `half + x^{n/2} half`. Each half can then be split down further and the middle factor `x^{n/2}` is always a square of the lower-level middle factors.

Starting from the bottom:

```
sum(2) = 1+x
sum(4) = sum(2) + x^2 sum(2)
sum(8) = sum(4) + x^4 sum(4)
...
```

Issue #55, opened 2018-04-16
Correct use of sponge in the Proof Transcript


oleganza:
The paper on [Duplex Keccak](https://keccak.team/files/SpongeDuplex.pdf) defines the duplex mode as absorbing and squeezing `<=r` (rate) bits (including padding) at each Keccak invocation. Our current implementation allows writing arbitrary string and then reading arbitrary string w/o absorbing any input, with padding applied only to the last block, and only at the absorbing phase.

<img width="877" alt="screen shot 2018-04-16 at 15 28 45" src="https://user-images.githubusercontent.com/698/38838464-09742416-418b-11e8-8c67-b478a87d0e49.png">

Fixing this would require a new Keccak API designed specifically to support duplex mode. Crate `tiny-keccak` does not seem to support this at the moment.

oleganza:
We are lucky that current version of `tiny-keccak` (1.3) actually allows doing exactly the duplex mode as described in the [Duplexing Keccak](https://keccak.team/files/SpongeDuplex.pdf) paper. Meaning, if we squeeze less that `rate` number of bytes, we will not disturb the internal state. Doing this right after absorbing gives us the duplex behavior that we need.

oleganza:
Seems like [STROBE](https://strobe.sourceforge.io/specs/#warn.sidechannel) does not length-prefix the input data to allow streaming, and instead relies on either user providing appropriate metadata inputs, or MACs. Same goes for MACs themselves: Mike has a notice saying:

> the first message that Alice sends to Bob will verify with a truncated MAC, but any message that either sends later in the protocol will fail, because of the different transcript before that point. This is unlikely to cause a huge security problem.

cathieyun:
What's the status of this issue, given merlin integration?

hdevalence:
#133 closes this issue.

PR #58, opened 2018-04-24
ProofTranscript: use stock `tiny-keccak` crate with duplex construction

oleganza:
This PR implements a Duplex Keccak mode (see #55 for details). We also switch back to non-forked `tiny-keccak` crate so we can publish the bulletproofs crate.

### DIFF
--- a/src/proof_transcript.rs
+++ b/src/proof_transcript.rs
@@ -27,12 +27,13 @@ use byteorder::{ByteOrder, LittleEndian};
 /// ensure that their challenge values are bound to the *entire* proof
 /// transcript, not just the sub-protocol.
 ///
-/// Internally, the `ProofTranscript` is supposed to use Keccak to
-/// absorb incoming messages and to squeeze challenges.  The
-/// construction currently used is ad-hoc, has no security analysis,
-/// and is **only suitable for testing**.
+/// ## Warning
 ///
-/// # Example
+/// Internally, the `ProofTranscript` uses ad-hoc duplex construction
+/// using Keccak that absorbs incoming messages and squeezes challenges.
+/// There is no security analysis yet, so it is **only suitable for testing**.
+///
+/// ## Example
 ///
 /// ```
 /// # extern crate curve25519_dalek;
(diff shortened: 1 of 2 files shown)


src/proof_transcript.rs
```diff
@@ -63,41 +59,49 @@ use byteorder::{ByteOrder, LittleEndian};
...
+            rate: 1600/8 - (2*128/8), // 168 bytes
+            write_offset: 0,
         };
+        // We will bump the version prefix each time we
+        // make a breaking change in order to disambiguate
+        // from the previous versions of this implementation.
+        ro.commit(b"ProofTranscript v2");
         ro.commit(label);
-        // makes sure the label is disambiguated from the rest of the messages.
-        ro.pad();
```
oleganza: We don't really need to pad here, as we don't pad/permute in-between multiple commitments. And `clone()` would work correctly with arbitrary offset.

src/proof_transcript.rs
```diff
@@ -130,12 +140,60 @@ impl ProofTranscript {
...
-    /// Pad separates the prior operations by padding
-    /// the rest of the block with zeroes and applying a permutation.
-    /// Each incoming message is length-prefixed anyway, but padding
-    /// enables pre-computing and re-using the oracle state.
-    fn pad(&mut self) {
+    /// Extracts a pair of challenge scalars.
+    ///
+    /// This is a convenience method that extracts 128 bytes and
+    /// reduces each 64-byte half modulo the group order.
+    pub fn challenge_scalars_pair(&mut self) -> (Scalar, Scalar) {
```
hdevalence: Do we really need a custom function for this? It seems kind of non-orthogonal. If an API client wants to get `n` scalars and doesn't want to repeatedly call `challenge_scalar`, they could extract the bytes they want themselves.

hdevalence:
LGTM

oleganza:
Note: this had a bug which was hot-fixed later here: https://github.com/chain/ristretto-bulletproofs/commit/28c6a8ae51a2e9567d134c3803de558bce8e8581

PR #59, opened 2018-04-24
Aggregated range proof notes

cathieyun:
Adds the "how do we get the aggregated proof checks" notes to the `notes` file. Does not explain the final mega_mult, or what how the dealer and party states work - I'm leaving that for another PR, under the `aggregated_range_proofs` documentation.


cathieyun:
Moving this to #64 so I don't have to deal with rebasing.

PR #61, opened 2018-04-26
Add a `SinglePartyAggregator` and refactor test code

hdevalence:
- renames `Proof` to `AggregatedProof`
- moves the orchestration logic from the tests into a convenience `SinglePartyAggregator`
- refactors the test code to perform a full round trip of create → serialize → deserialize → verify

### DIFF
--- a/src/aggregated_range_proof/mod.rs
+++ b/src/aggregated_range_proof/mod.rs
@@ -76,117 +76,112 @@ impl SinglePartyAggregator {
 
 #[cfg(test)]
 mod tests {
-    use std::iter;
+    use rand::OsRng;
 
-    use rand::{OsRng, Rng};
+    use super::*;
 
-    use curve25519_dalek::scalar::Scalar;
-    use proof_transcript::ProofTranscript;
+    use generators::PedersenGenerators;
 
-    use super::dealer::*;
-    use super::messages::*;
-    use super::party::*;
+    /// Given a bitsize `n`, test the following:
+    ///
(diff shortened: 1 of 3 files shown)


src/aggregated_range_proof/mod.rs
```diff
@@ -1,122 +1,187 @@
...
+    fn generate_proof<R: Rng>(
+        generators: &Generators,
+        transcript: &mut ProofTranscript,
         rng: &mut R,
-        values: Vec<u64>,
+        values: &[u64],
         n: usize,
-    ) -> (Proof, Vec<ProofShareVerifier>) {
-        use generators::{Generators, PedersenGenerators};
+    ) -> Result<AggregatedProof, &'static str> {
```
oleganza: Can it actually ever fail? Maybe we can do `unwrap()` inside the implementation to avoid exposing unused errors?

src/aggregated_range_proof/mod.rs
```diff
@@ -1,122 +1,187 @@
...
+
+        // Prover's scope
+        {
+            // 1. Generate the proof
+
+            let mut rng = OsRng::new().unwrap();
+            let mut transcript = ProofTranscript::new(b"AggregatedRangeProofTest");
+
+            // XXX this takes max = 2^{n-1} to avoid problems at n = 64
+            // would be better to use max = 2^n - 1
```
oleganza: maybe use u128 and cast to u64 after subtracting one?

PR #62, opened 2018-04-26
Generalize fast sum of powers for any length, not just power-of-two

oleganza:


hdevalence:
Is there a reason to do this instead of using the identity `1 + ... + x^{n-1} = (1 - x^n) / (1 - x)`?

oleganza:
Doesn't one inversion cost ≈200 multiplications? This implementation does `(2..3)*lg n` multiplications (`2*lg(n)` when `n` is power of two).

oleganza:
But i did not know about `1 + ... + x^{n-1} = (1 - x^n) / (1 - x)`, thanks for pointing it out!

hdevalence:
Cool, I'm convinced this works but I'm not sure I understand why yet, so I'm working on some notes right now.

hdevalence:
The power-of-two recurrence is really nicely explained and it makes sense, but I don't understand why the recurrence for general `n` works. Is it possible to prove it?

oleganza:
I just rebased this on `main`, for good hygiene, not with an expectation to merge this soon.

In fact, only RP code require this helper, and in RP the input is always a power-of-two. For circuits, the `delta` is defined as an inner product of powers of a challenge with circuit-specific weights, so this whole function is not relevant.

Issue #60, opened 2018-04-25
Dealer in aggregated range proof implementation should verify shares


hdevalence:
Moving from https://github.com/chain/ristretto-bulletproofs/pull/57#discussion_r184209569

The dealer logic should verify proof shares internally.

hdevalence:
I think there's a slight complication.

The ideal flow would be that the dealer decides whether or not the proof shares are trusted (e.g., self-aggregation). If the shares are untrusted, the dealer should

1. Return `Ok(proof)` if the aggregated proof verifies
2. Otherwise, check each share for validity and return an `Err` with info about which shares were invalid.

The reason to do this is that verifying the aggregated proof is cheaper than verifying each share individually.  However, the dealer can't just call `.verify()` since that code attempts to replay the proof and generate challenges, which requires having the starting transcript state, which the dealer already destroyed by running the protocol.

One option would be to save the transcript state at the beginning, then pass it to the `verify` call. Another would be to duplicate the verification logic inside of the dealer, to take account of the fact that the dealer still has all of the challenges at hand. The second option seems less nice.

oleganza:
> One option would be to save the transcript state at the beginning

Sounds right to me. Cloning a transcript is a pretty cheap operation and we already support that. (Although, we should probably add tests for that use case.)

hdevalence:
Maybe this would also be good to do for parameters like the `GeneratorsView` that's used at the very end? i.e., pass all of them into `new`, so that once the initial `Dealer` state is created, the only inputs needed to advance the state are the protocol messages.

@cathieyun WDYT?

cathieyun:
I do think it's better to have the transcript passed between the dealer states, instead of being passed in every time (the counterargument for that is that the rng is passed in every time... but that's different in that it doesn't matter if anyone touches the rng in between dealer states). 

with regards to challenges - is it possible to (instead of cloning the transcript at the beginning) make a new transcript with the same initial string, and the use that for the verification?

hdevalence:
I think we can't make a new transcript except by cloning the initial one, since there's no way for the dealer code to know the customization string -- or even if the state was created by using a customization string, since the RP protocol could be composed into a larger protocol (in the same way that the IPP is composed into the RP protocol).

I'll patch up the dealer states to hold on to their arguments.

PR #63, opened 2018-04-26
Keep dealer parameters internal to the dealer state

hdevalence:
(Follows #61, prelude to #60)

Moves the transcript state inside the dealer state. The reasoning is in the commit messages and in this thread: https://github.com/chain/ristretto-bulletproofs/issues/60#issuecomment-384742059

### DIFF
--- a/src/aggregated_range_proof/dealer.rs
+++ b/src/aggregated_range_proof/dealer.rs
@@ -13,11 +13,12 @@ pub struct Dealer {}
 
 impl Dealer {
     /// Creates a new dealer coordinating `m` parties proving `n`-bit ranges.
-    pub fn new<'a>(
+    pub fn new<'a, 'b>(
+        gens: GeneratorsView<'b>,
         n: usize,
         m: usize,
         transcript: &'a mut ProofTranscript,
-    ) -> Result<DealerAwaitingValueCommitments<'a>, &'static str> {
+    ) -> Result<DealerAwaitingValueCommitments<'a, 'b>, &'static str> {
         if !n.is_power_of_two() || n > 64 {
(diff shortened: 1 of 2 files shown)


hdevalence:
(note: this branch follows from #61, which isn't merged yet, so GitHub shows it as including those commits. I made seperate PRs so that each change is smaller).

src/aggregated_range_proof/dealer.rs
```diff
@@ -13,11 +13,12 @@ pub struct Dealer {}
...
     /// Creates a new dealer coordinating `m` parties proving `n`-bit ranges.
-    pub fn new(
+    pub fn new<'a, 'b>(
+        gens: GeneratorsView<'b>,
         n: usize,
         m: usize,
-        transcript: &mut ProofTranscript,
-    ) -> Result<DealerAwaitingValueCommitments, &'static str> {
+        transcript: &'a mut ProofTranscript,
+    ) -> Result<DealerAwaitingValueCommitments<'a, 'b>, &'static str> {
```
oleganza: Does this mean we cannot serialize the dealer? Seems a bit limiting if the dealer object must remain in memory tied to a lifetime of its parameters.

oleganza: What if you move the transcript in, and then return it at the end, so the outer protocol can continue using it?

hdevalence: Yes, this means you cannot serialize the dealer. I think that's fine, since the dealer is created once per run of the (online) aggregation protocol, and I don't think there's a good reason to stop in the middle of the protocol, save the state, and then resume it later -- you could just start over if you needed to do that.

cathieyun: makes sense, 👍

PR #64, opened 2018-04-26
Aggregated range proof notes

cathieyun:
Adds the "how do we get the aggregated proof check equations" notes to the `notes` file. Does not explain the final `mega_mult`, or what how the dealer and party states work - I'm leaving that for another PR, under the `aggregated_range_proofs` documentation.

Moved changes from #59 to avoid rebasing headaches.
Most notes based originally on the latex notes from the research repo: https://github.com/chain/research/pull/16


cathieyun:
I'm pretty proud of these notes :) PTAL @hdevalence & @oleganza ?

hdevalence:
These are good. I have some general comments below and then I'll write some inline comments for little things.

1. I think it would be good to expand the introductory paragraph to give more context, so that the reader gets more of a structure upfront that they can fit the rest of the explanation into. Specifically, I think it would be good to mention:
  - that the aggregation protocol is an MPC protocol;
  - that the Bulletproofs paper sketches two versions of MPC aggregation, but we only implement one of them (and why we picked the one we did);
  - that the players in the protocol are a single dealer and `m` parties, who don't reveal their secrets to each other;
  - maybe a rough outline of how the parts of the parts of the non-aggregated protocol are split over the players in the protocol (?)

2. Sometimes it's not immediate which parts are the same as in the non-aggregated case and which parts are different. For instance, the section about "proving range statements with bit vectors" is the same as in the non-aggregated case, except for a subscript, but the reader would have to check this.

3. The explanation of how the offsets give 'views' into a giant vector of powers of the challenges is hinted at towards the end ("therefore, we can simplify the following statements") but it might be helpful to do this more explicitly and more up-front. Maybe this could be fit into the part where the offsets are introduced?

4. In the "proving that t_0" is correct section, the statements about each party's values are combined by summing them. But elsewhere, when we combine statements, we have to add random factors. I think it's important to have an explanation of why, in this part, it's OK to just sum the statements rather than take a random linear combination. 

5. It would be good to describe how to detect whether a single party's share was malformed, and describe how the dealer can verify that the protocol was performed correctly (giving a valid proof) or else detect which party's shares were malformed.

docs/notes.md
```diff
@@ -603,4 +603,242 @@ sends a pair \\((L\_j,R\_j)\\) of points at each step \\(j = k\dots1\\). An
...
+-----------------------------------------
+
+Let \\({\mathbf{a}}\_{Lj}\\) be the vector of bits of \\(v_j\\).
+Then \\(v_j\\) can be represented as:
+\\[
+\begin{aligned}
+  v_j &= {\langle {\mathbf{a}}\_{Lj}, {\mathbf{2}}^{n} \rangle}
+\end{aligned}
+\\]
+We need \\({\mathbf{a}}\_{Lj}\\) to be a vector of integers \\(\\{0,1\\}\\).
```
hdevalence: Notation: I think `_{Lj}` is confusing. I'm not 100% sure of the notationally-best way to write the party index.

Superscripts would have to have `(j)` or something to distinguish them from exponents, but we're already overloading exponents with the exponential-vector thing.

Subscripts don't have the same confusion, but we're already using subscripts. Maybe the least awkward thing would be to use comma separated subscripts, like `a_{L, j}`?

docs/notes.md
```diff
@@ -603,4 +603,242 @@ sends a pair \\((L\_j,R\_j)\\) of points at each step \\(j = k\dots1\\). An
...
+\\({\mathbf{l}}\_j(x)\\), \\({\mathbf{r}}\_j(x)\\) are correctly formed, and that
+\\(t_j(x) = {\langle {\mathbf{l}}\_j(x), {\mathbf{r}}\_j(x) \rangle}\\).
+
+We can combine the statements about \\(t_j(x)\\), \\({\mathbf{l}}\_j(x)\\), and \\({\mathbf{r}}\_j(x)\\) from all `m` parties in the following manner:
+
+\\[
+\begin{aligned}
+  t(x) &= \sum_{j=0}^{m-1} t_j(x)\\\\
+  {\mathbf{l}}(x) &= {\mathbf{l}}\_{j=0}(x) || {\mathbf{l}}\_{j=1}(x) || \dots || {\mathbf{l}}\_{j=m-1}(x) \\\\
+  {\mathbf{r}}(x) &= {\mathbf{r}}\_{j=0}(x) || {\mathbf{r}}\_{j=1}(x) || \dots || {\mathbf{r}}\_{j=m-1}(x) \\\\
```
hdevalence: This should maybe be
``` 
 {\mathbf{r}}(x) &= {\mathbf{r}}\_{0}(x) || {\mathbf{r}}\_{1}(x) || \dots || {\mathbf{r}}\_{m-1}(x) \\\\ 
```
instead of 
```
{\mathbf{r}}(x) &= {\mathbf{r}}\_{j=0}(x) || {\mathbf{r}}\_{j=1}(x) || \dots || {\mathbf{r}}\_{j=m-1}(x) \\\\ 
```

docs/notes.md
```diff
@@ -603,4 +603,242 @@ sends a pair \\((L\_j,R\_j)\\) of points at each step \\(j = k\dots1\\). An
...
+
+The prover chooses vectors of blinding factors
+\\[
+{\mathbf{s}}\_{Lj}, {\mathbf{s}}\_{Rj} \\;{\xleftarrow{\\$}}\\; {\mathbb Z\_p}^{n}
+\\]
+and uses them to construct blinded vector polynomials:
+\\[
+\begin{aligned}
+  {\mathbf{l}}\_j(x) &= ({\mathbf{a}}\_{Lj} + {\mathbf{s}}\_{Lj} x) - z {\mathbf{1}} & \in {\mathbb Z\_p}[x]^{n}  \\\\
+  {\mathbf{r}}\_j(x) &= {\mathbf{y}}^{n} \cdot y_j \circ \left( ({\mathbf{a}}\_{Rj} + {\mathbf{s}}\_{Rj} x\right)  + z {\mathbf{1}}) + z^{2} z_j {\mathbf{2}}^{n} &\in {\mathbb Z\_p}[x]^{n}
```
hdevalence: Previously, `y_j = y^j` was introduced as a new variable (I guess the intent is to distinguish its role as a per-party offset, rather than its construction as `y^j` ??). In that case, does it make sense to *also* introduce `\mathbf{y}_j` as the per-party view into the larger exponent vector, for the same reasons?

docs/notes.md
cathieyun: makes sense - I was looking for a better way to do this, and `a_{L, j}` is better than the current way.

docs/notes.md
cathieyun: hm, I was trying to find some way to differentiate `r_{j=0}` from the `r_0` and `r_1` values that are the exponents of the `r` polynomial (used in lines 692-697)... I think that this is clearer that it is the `j` value that is `0`, but I think the subscript overloading is inherently clunky :/

docs/notes.md
cathieyun: oo that's a great idea!

cathieyun: and it was introduced as `y_j = y^(j*n)` so readers wouldn't have to worry about the exponent, just that it's the `j`th party's share of `y`. But your suggestion is even better / clearer than that!

docs/notes.md
hdevalence: Ah, I see what you mean... you're right, this is awkward. Maybe we can think of a better/tweaked notation?

cathieyun: what about `r_(j)` for `j`-party subscripts in general, and in this case it would be `r_(0)`, `r_(1)` which is distinct from `r_0`, `r_1`?

hdevalence: That seems reasonable! It is still a bit awkward, but I'm at a loss to think of something that's better.

cathieyun:
Updated notes from feedback: 
- [x] use `_(j)` instead of `_j` notation for per-party shares, to prevent confusion with `t_0, t_1`, etc.
- [x] better explanation in the intro (see hdevalence's comment for more details)
- [x] make it clear which parts are the same as single-range proof (by just saying so & linking to that section), to prevent duplicate explanations and to make things clearer.
- [x] explicit explanation of how offsets give 'views' into a giant vector of powers
- [x] explain why it's okay to sum in certain parts & have to concatenate in other parts
- [ ] description of checking of single party's share was malformed -> I'm thinking of doing this in the implementation notes instead, WDYT?
- [x] use `\vec y_(j)` vector (instead of `\vec y^n * y^(n*j)`) - clearer

hdevalence:
:eyes:

docs/notes.md
```diff
@@ -603,4 +603,254 @@ sends a pair \\((L\_j,R\_j)\\) of points at each step \\(j = k\dots1\\). An
 additional and final step involves sending a pair of scalars
 \\((a^{(0)}\_0,b^{(0)}\_0)\\) and checking the final relation directly.
 
-[bulletproofs_paper]: https://eprint.iacr.org/2017/1066.pdf
+Aggregated Range Proof
+======================
+
+We want to create an aggregated range proof for `m` values that is more efficient to create and verify than `m` individual range proofs.
```
hdevalence: Maybe this could lead by saying that the goal is to take advantage of the logarithmic size of the inner-product protocol?

As is I think it isn't quite right, because creating the aggregated range proof isn't any more efficient than individual proofs, and I'm not sure how verifying an aggregated proof compares to batch-verifying individual proofs. But the size is a win from aggregation.

hdevalence: nit: `` `m` `` should be `\\(m\\)` so that it's math instead of code

docs/notes.md
```diff
@@ -603,4 +603,254 @@ sends a pair \\((L\_j,R\_j)\\) of points at each step \\(j = k\dots1\\). An
...
+
+The aggregation protocol is a multi-party computation protocol, involving `m` parties (one party per value) and one dealer, where the parties don't reveal their secrets to each other. The parties share their commitments with the dealer, and the dealer generates and returns challenge variables. The parties then share their proof shares with the dealer, and the dealer combines their shares to create an aggregated proof. 
+
+The Bulletproofs paper outlines two versions of multi-party computation aggregation - one with a constant number of rounds but communication that is linear in both `m` and the binary encoding of the range, and one with a logarithmic number of rounds and communication that is only linear in `m`. We chose to implement the first version because the two versions don't differ significantly in proof size, and the first version is more straightforward.
```
hdevalence: I think it's better not to say "don't differ significantly" because resulting proof sizes are exactly the same.

Rather than giving the asymptotics on the number of rounds and size of communication, we could say that the difference is whether the inner-product proof is performed by the dealer (which requires sending the vectors used for the inner-product), or whether the inner-product proof is performed using MPC (which sends less data, but requires one round for each iteration of the inner-product protocol).

From this point of view I think it's a lot more clear why the choice we made is the right one, because even for 64-bit ranges each party's vectors are only a few KB, which isn't worth doing extra round trips.

docs/notes.md
```diff
@@ -603,4 +603,254 @@ sends a pair \\((L\_j,R\_j)\\) of points at each step \\(j = k\dots1\\). An
...
+
+For more information on how the aggregation protocol works and is implemented, see the [protocol notes](../aggregated_range_proof/index.html). 
+
+The aggregated range proof has the same form as the individual range proof, in that the provers (the parties) still perform the same calculations to prove that \\(t(x) = \langle \mathbf{l}(x), \mathbf{r}(x) \rangle \\) and that \\(t_0, \mathbf{l}(x), \mathbf{r}(x)\\) are correct. The difference is that the challenge values are obtained from the dealer, which generates them by combining commitments from all the parties, and that the calculations of different parties are seperated by different powers of the challenge scalars \\(y\\) and \\(z\\).
+
+We will explain how one piece of the aggregated proof is generated for party `j`, and then will show how all of the pieces for all of the `m` parties can be combined into one aggregated proof.
```
hdevalence: nit: `` `j` `` should be `\\(j\\)`

docs/notes.md
```diff
@@ -603,4 +603,254 @@ sends a pair \\((L\_j,R\_j)\\) of points at each step \\(j = k\dots1\\). An
...
+
+\\[
+\begin{aligned}
+  t(x) &= \sum_{j=0}^{m-1} t\_{(j)}(x)\\\\
+  {\mathbf{l}}(x) &= {\mathbf{l}}\_{(0)}(x) || {\mathbf{l}}\_{(1)}(x) || \dots || {\mathbf{l}}\_{(m-1)}(x) \\\\
+  {\mathbf{r}}(x) &= {\mathbf{r}}\_{(0)}(x) || {\mathbf{r}}\_{(1)}(x) || \dots || {\mathbf{r}}\_{(m-1)}(x) \\\\
+\end{aligned}
+\\]
+
+We can add the \\(t_{(j)}(x)\\) values together to create \\(t(x)\\) instead of taking a linear combination of \\(t_{(j)}(x)\\) values, because each \\(t_{(j)}(x)\\) is calculated with the \\(\mathbf{y}^n\_{(j)}\\) and \\(z_{(j)}\\) challenge variables that are unique to that party`j`, so all of the \\(t_{(j)}(x)\\) values will be offset from one another.
```
hdevalence: should be "random linear combination" rather than "linear combination" since adding together is also a linear combination

docs/notes.md
```diff
@@ -603,4 +603,254 @@ sends a pair \\((L\_j,R\_j)\\) of points at each step \\(j = k\dots1\\). An
...
+
+If we combine all of the statements about \\(t\_{(j)0}\\) from all of the `j` parties by adding them together, then we get:
+
+\\[
+\begin{aligned}
+  \sum_{j=0}^{m-1}t_{(j)}(x) B + \sum_{j=0}^{m-1}{\tilde{t}}\_{(j)}(x) {\widetilde{B}} \stackrel{?}{=} z^2 \sum_{j=0}^{m-1} z_{(j)} V_{(j)} + \sum_{j=0}^{m-1} \delta_{(j)}(y,z) B + x \sum_{j=0}^{m-1} T\_{(j)1} + x^{2} \sum_{j=0}^{m-1} T\_{(j)2}
+\end{aligned}
+\\]
+
+We can combine the values and commitments by summing them directly. We can do this instead of having to take a linear combination, because each party's values and commitments are already offset by the values \\(\mathbf{y}^n\_{(j)}\\) and \\(z_{(j)}\\) that are unique to that party.
```
hdevalence: "random"

docs/notes.md
```diff
@@ -603,4 +603,254 @@ sends a pair \\((L\_j,R\_j)\\) of points at each step \\(j = k\dots1\\). An
...
+
+\\[
+\begin{aligned}
+  t(x) &= \sum_{j=0}^{m-1} t\_{(j)}(x)\\\\
+  {\tilde{t}}(x) &= \sum_{j=0}^{m-1}{\tilde{t}}\_{(j)}(x)\\\\
+  T_1 &= \sum_{j=0}^{m-1} T_{(j)1}\\\\
+  T_2 &= \sum_{j=0}^{m-1} T_{(j)2}\\\\
+  \delta(y,z) &= \sum_{j=0}^{m-1} \delta\_{(j)}(y,z)\\\\
```
hdevalence: In the rendered docs this is a little confusing to read, because each of the previous lines are disjoint statements, while this last one has some simplifications. Maybe something like

```
The aggregated values and commitments are
\\[
... stuff ..
\\]
and the expression for \\(\delta(y,z)\\) becomes
\\[
.. delta stuff. ...
\\]
```
so that there's a break in the writing?

docs/notes.md
cathieyun: 👍

docs/notes.md
cathieyun: 👍

docs/notes.md
```diff
@@ -603,4 +603,254 @@ sends a pair \\((L\_j,R\_j)\\) of points at each step \\(j = k\dots1\\). An
...
+\begin{aligned}
+  t(x) B + {\tilde{t}}(x) {\widetilde{B}} \stackrel{?}{=} \sum_{j=0}^{m-1} z^{j+2} V_{(j)} + \delta(y,z) B + x T\_{1} + x^{2} T\_{2},\\\\
+  \delta(y,z) = (z - z^{2}) \cdot {\langle {\mathbf{1}}, {\mathbf{y}}^{n \cdot m} \rangle} - \sum_{j=0}^{m-1} z^{j+3} \cdot {\langle {\mathbf{1}}, {\mathbf{2}}^{n \cdot m} \rangle}\\\\
+\end{aligned}
+\\]
+
+Proving that \\({\mathbf{l}}(x)\\), \\({\mathbf{r}}(x)\\) are correct
+---------------------------------------------------------------------
+
+Proving that \\({\mathbf{l}}\_{(j)}(x)\\), \\({\mathbf{r}}\_{(j)}(x)\\) are correct requires first creating commitments to the variables, and then proving a relation over the commitments. For an explanation of how the commitments are created and how the relation is derived, see the [proving that \\({\mathbf{l}}(x)\\), \\({\mathbf{r}}(x)\\) are correct](index.html#proving-that-mathbflx-mathbfrx-are-correct) step of the single-value range proof. The relation over the commitments to prove is:
```
hdevalence: "The relation over the commitments to prove" is a little awkward, maybe "The statement each party wants to prove is" (since I think (from the subscripts) that this is a per-party statement that hasn't been combined yet?)

docs/notes.md
```diff
@@ -603,4 +603,254 @@ sends a pair \\((L\_j,R\_j)\\) of points at each step \\(j = k\dots1\\). An
...
+  A &= \sum_{j=0}^{m-1} A_{(j)} \\\\
+  S &= \sum_{j=0}^{m-1} S_{(j)} \\\\
+\end{aligned}
+\\]
+
+With these observations, we can simplify the combined `m`-party statement about \\({\mathbf{l}}(x)\\) and \\({\mathbf{r}}(x)\\) into:
+
+\\[
+\begin{aligned}
+  {\langle {\mathbf{l}}(x), {\mathbf{G}} \rangle} + {\langle {\mathbf{r}}(x) \circ {\mathbf{y}^{-n \cdot m}} , {\mathbf{H}} \rangle} \stackrel{?}{=} -{\widetilde{e}} {\widetilde{B}} + A + x S - z{\langle {\mathbf{1}}, {\mathbf{G}} \rangle} + z{\langle {\mathbf{1}}, {\mathbf{H}} \rangle} + \sum_{j=0}^{m-1} {\langle z^{j+2} \cdot (\mathbf{y}^{n \cdot m}\_{[j \cdot n : (j+1) \cdot n - 1]})^{-1} \circ {\mathbf{2}}^n, {\mathbf{H}}\_{[j \cdot n : (j+1) \cdot n - 1]} \rangle}
```
hdevalence: The H' terms
```
(\mathbf{y}^{n \cdot m}\_{[j \cdot n : (j+1) \cdot n - 1]})^{-1} \circ {\mathbf{2}}^n
```
are pretty inscrutable (at least to me), but the way you described it in the code seems more clear, maybe it's possible to factor out this part into a new vector that's built up in the same way as in the code?

docs/notes.md
```diff
@@ -603,4 +603,254 @@ sends a pair \\((L\_j,R\_j)\\) of points at each step \\(j = k\dots1\\). An
...
+
+For more information on how the aggregation protocol works and is implemented, see the [protocol notes](../aggregated_range_proof/index.html). 
+
+The aggregated range proof has the same form as the individual range proof, in that the provers (the parties) still perform the same calculations to prove that \\(t(x) = \langle \mathbf{l}(x), \mathbf{r}(x) \rangle \\) and that \\(t_0, \mathbf{l}(x), \mathbf{r}(x)\\) are correct. The difference is that the challenge values are obtained from the dealer, which generates them by combining commitments from all the parties, and that the calculations of different parties are seperated by different powers of the challenge scalars \\(y\\) and \\(z\\).
+
+We will explain how one piece of the aggregated proof is generated for party `j`, and then will show how all of the pieces for all of the `m` parties can be combined into one aggregated proof.
+
+Party `j` begins with a secret value \\(v_{(j)}\\), and wishes to convince the verifier that \\(v_{(j)} \in [0, 2^n)\\) without revealing \\(v_{(j)}\\). We use the notation that the subscript \\({(j)}\\) denotes the `j`th party's value.
```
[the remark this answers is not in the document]
cathieyun: 👍

docs/notes.md
cathieyun: 👍

docs/notes.md
cathieyun: I'm confused where you mean by "in the code" - in the inner-product-protocol documentation? Or in the comments in the code? (I don't think I wrote any extensive comments explaining the math in the code)

Either way, I'm guessing what you mean is to use the `H'` vector instead of expanding that to `H` and the `y` exponent - I'll give that a shot!

cathieyun: It looks much clearer, I like it! 
👍

docs/notes.md
```diff
@@ -603,4 +603,254 @@ sends a pair \\((L\_j,R\_j)\\) of points at each step \\(j = k\dots1\\). An
...
+\end{aligned}
+\\]
+
+Proving vectors of statements with a single statement
+-----------------------------------------------------
+
+We want to combine the above three statements into a single statement for party `j`, as in the [proving vectors of statements](index.html#proving-vectors-of-statements-with-a-single-statement) step of the single-value range proof. We will additionally introduce challenge values \\(\mathbf{y}^n\_{(j)}\\) and \\(z_{(j)}\\) that are unique to each party `j`. Since these challenge values are independent for each party, we can later merge the per-party combined statements into one statement for all `m` parties.
+
+First, we will combine each of the two vector-statements into a single statement using the verifier's choice of challenge value \\(y\\) that is shared across all parties
+, and offset by vector \\(\mathbf{y}^n\_{(j)} = \mathbf{y}^{n \cdot m}\_{[j \cdot n : (j+1) \cdot n - 1]} \\), a length `n` slice into vector \\(\mathbf{y}^{n \cdot m}\\) that is unique to each party `j`:
```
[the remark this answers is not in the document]
cathieyun: 👍

docs/notes.md
cathieyun: 👍

PR #67, opened 2018-04-27
Have the dealer verify shares

### LINKED ISSUE #60: Dealer in aggregated range proof implementation should verify shares

hdevalence:
This closes #60.

Right now the information about which shares are invalid isn't sent upwards, because we don't have any custom error types, but as soon as we do that (using `failure`), we will be able to surface which shares are invalid.

For now it just gives a generic error and prints the bad shares to stdout.


src/aggregated_range_proof/mod.rs
```diff
@@ -180,4 +177,72 @@ mod tests {
...
+        // Parties 0, 2 are honest and use a 32-bit value
+        let v0 = rng.next_u32() as u64;
+        let v0_blinding = Scalar::random(&mut rng);
+        let party0 = Party::new(v0, v0_blinding, n, &generators).unwrap();
+
+        let v2 = rng.next_u32() as u64;
+        let v2_blinding = Scalar::random(&mut rng);
+        let party2 = Party::new(v2, v2_blinding, n, &generators).unwrap();
+
+        // Parties 1, 3 are dishonest and use a 32-bit value
```
oleganza: typo, must be "use a 64-bit value".

PR #72, opened 2018-04-27
Add aggregate proof benchmarks

hdevalence:
The creation and verification benchmarks are done using Criterion's `bench_function_over_inputs` to run over a range of aggregation sizes, currently `m= 1,2,4,8,16,32`.

It is probably worth investigating the aggregate RP performance, since it seems slower, and also linear in the aggregation size (????):

```ascii
verify_rangeproof_n_32  time:   [674.08 us 674.33 us 674.60 us]

Aggregated 32-bit rangeproof verification/1                                                                             
                        time:   [2.2918 ms 2.2921 ms 2.2924 ms]
Aggregated 32-bit rangeproof verification/2                                                                             
                        time:   [4.2642 ms 4.2657 ms 4.2674 ms]
Aggregated 32-bit rangeproof verification/4                                                                             
                        time:   [8.1347 ms 8.1354 ms 8.1363 ms]
Aggregated 32-bit rangeproof verification/8                                                                             
                        time:   [15.947 ms 15.948 ms 15.949 ms]
```

### DIFF
--- a/benches/bulletproofs.rs
+++ b/benches/bulletproofs.rs
@@ -112,4 +112,97 @@ criterion_group!{
     targets = verify_rp_8, verify_rp_16, verify_rp_32, verify_rp_64
 }
 
-criterion_main!(create_rp, verify_rp);
+static AGGREGATION_SIZES: [usize; 6] = [1, 2, 4, 8, 16, 32];
+
+fn create_aggregated_rangeproof_helper(n: usize, c: &mut Criterion) {
+    use ristretto_bulletproofs::aggregated_range_proof::SinglePartyAggregator;
+
+    let label = format!("Aggregated {}-bit rangeproof creation", n);
+
+    c.bench_function_over_inputs(
+        &label,
+        move |b, &&m| {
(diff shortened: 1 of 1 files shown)


benches/bulletproofs.rs
```diff
@@ -112,4 +112,97 @@ criterion_group!{
...
+                &generators,
+                &mut transcript,
+                &mut rng,
+                &values,
+                n,
+            ).unwrap();
+
+            b.iter(|| {
+                // Each proof creation requires a clean transcript.
+                let mut transcript = ProofTranscript::new(b"AggregateRangeProofBenchmark");
```
oleganza: I know that the overhead of creating a new transcript is very small, but it's a bit cleaner to clone an existing transcript here, than create one from scratch.

oleganza:
I think it is expected that the performance is linear. Multiscalar multiplication with Straus gets linear around 100-200 points and our multi-proof aggregation is way past it.

hdevalence:
I think the problem is that the proof verification function doesn't take generators as a parameter, so it recomputes them all every time.

PR #73, opened 2018-04-27
Add aggregated proof implementation notes

cathieyun:
Add internal documentation on how the aggregated proof was implemented, in the `aggregated-proof-protocol` file. 

(This is distinct from the "aggregated proof notes", PR #64, which explains the math behind the aggregated proof calculations, but not how it is implemented).

- [x] prover's algorithm
- [x] verifier's algorithm
- [x] state machines for party and dealer
- [x] how individual share validation works

After I merge this in, I'll make another PR do the step to overwrite the existing rangeproof notes with these notes.

I am particularly looking for feedback on the "Prover and verifier state machines", because currently the wording is not very clear or precise.


cathieyun:
PTAL - I am particularly looking for feedback on the "Prover and verifier state machines", because currently the wording is not very clear or precise.

hdevalence:
The source tree rearrangement in #84 moved things around from under this changeset, so I'll try rebasing it onto the current `main` before reviewing it.

hdevalence:
Since the `aggregated_range_proof` module no longer exists (the single RP is a special case of the agg RP, and it's all the `range_proof` module), it seems better to overwrite the existing notes as part of this PR, since that way it's also easier to make sure that these notes don't miss anything that was in the previous ones.

docs/range-proof-protocol.md
```diff
@@ -1,190 +1,315 @@
-The `range_proof` module contains API for producing a range proof for a given integer value.
+The `aggregated_range_proof` module contains an API for producing a proof that multiple integer values are within a certain range.
```
hdevalence: This should probably refer to `range_proof` now that the `aggregated_range_proof` module is gone.

docs/range-proof-protocol.md
```diff
@@ -1,190 +1,315 @@
...
+This is a documentation for the internal implementation of the aggregated range proof. You may find the introduction to all the pieces of the range proof protocol in the [notes](../notes/index.html) module.
 
-The range proof is a zero-knowledge proof of the following relation
+The aggregated range proof is a zero-knowledge proof of the following relation:
 \\[
 \operatorname{ZK-PK}\left\\{
-  v \in {\mathbb Z\_p}
-  : v \in [0, 2^n)
+  v_{(j)} \in {\mathbb Z\_p}
+  : v_{(j)} \in [0, 2^n)
```
hdevalence: I think this might be clearer if it were written as `v_{(1)}, \ldots, v_{(m)}` since that reinforces that there are multiple values and also introduces the `m` parameter up front in the problem statement as the number of values.

docs/range-proof-protocol.md
```diff
@@ -1,190 +1,315 @@
...
-\\] where \\(n = 2^{k}\\) is a power of \\(2\\).
+\\] 
 
-Prover’s algorithm
+for \\(m\\) values \\(v_{(j)}\\), where \\(n\\) and \\(m\\) are both a power of \\(2\\).
+
+API overview
+------------
+
+The prover is composed of \\(m\\) individual parties and one dealer that coordinates messages and provides challenges.
```
hdevalence: I think it might be good not to use the word "prover" at all in this write-up, because the proving procedure is now split across all of the participants in the protocol, and it would be good to always be specific about what is done by whom.

docs/range-proof-protocol.md
```diff
@@ -1,190 +1,315 @@
...
+
+First, parties are instantiated with their secret values and the range for the range proof. The dealer assigns a unique index \\(j\\) to each party.
```
hdevalence: Rather than using the passive-voice "are instantiated with" (instantiated by whom?), this section would be a good place to explicitly describe the setting for the API: the protocol could be run locally, or across a network, and this API is supposed to be used by whatever code is orchestrating the protocol execution.

docs/range-proof-protocol.md
```diff
@@ -1,190 +1,315 @@
...
+
+Next, each party commits to its secret value, bits, and blinding factors via \\(V_{(j)}, A_{(j)}, S_{(j)}\\) as `ValueCommitment` and sends `ValueCommitment` to the dealer. The dealer receives and combines all of the `ValueCommitment`s from all of the parties, generates challenges \\(y,z\\) as `ValueChallenge`, and sends `ValueChallenge` to all of the parties.
```
hdevalence: This could be shortened a little bit by saying, e.g.,
```
... blinding factors via \\( ... \\) and sends this data in a `ValueCommitment` message to ...
```
which avoids saying the message name twice, and makes it clear that it is a message.

hdevalence:
I left a few inline comments, but here are some general ones about your state machine question:

- The structure of the code into `dealer` / `party` / `messages` modules maps onto the structure of the protocol, so it would be good to mention in the API description.

- We should think about the audiences for the internal and external docs and figure out what parts are relevant for each. In particular, the current API structure re-exports all of the `dealer` / `party` / `messages` modules as part of an externally-visible `aggregation` module, in order to allow people to do the MPC protocol over a network, or whatever is useful for them. But to actually use those pieces, you have to understand the protocol structure.  So maybe the image with the state machine description, and the explanation of the prover states would be better there? (Or maybe the image should go in both places?)

- Is it possible to export the diagram as an SVG instead of a PNG?

- I think it makes more sense to do the description of the types and the protocol flow after the diagram, and in these docs we don't need to refer to "the Rust type system" since they're already Rust docs, we can just say something like "each protocol state is encoded as a distinct type, and we use move semantics to ensure that a previous state is consumed as it transitions to the next state."

- We should mention that typechecking an instantiation of the MPC protocol ensures that the implementation is invulnerable to replay attacks.

oleganza:
I can work on the diagram in a proper editor to make a nice SVG.

docs/range-proof-protocol.md
cathieyun: 👍

docs/range-proof-protocol.md
cathieyun: 👍

docs/range-proof-protocol.md
cathieyun: 👍

oleganza:
Added SVG diagram and its source in a [Sketch](https://www.sketchapp.com) format. Also used the opportunity to add proper rounding to the bended lines.

screen shot 2018-05-07 at 13 34 29

docs/range-proof-protocol.md
cathieyun: 👍

docs/aggregation-api.md
```diff
@@ -0,0 +1,14 @@
+API for the aggregated multiparty computation protocol
```
hdevalence: It's probably good to have at least one line of text before this, since Rustdoc will use it as a summary line. If the summary line is a heading, then the heading will be inserted into other places in the docs where a heading doesn't look great.

docs/aggregation-api.md
```diff
@@ -0,0 +1,14 @@
...
....45.882 1.332.882zm7.992 1.26v-1.512h1.962v-4.914h-1.962v-1.53h3.636v1.8c.342-.882 1.134-1.98 2.7-1.98.594 0 1.134.144 1.494.342v1.944c-.396-.234-1.008-.378-1.53-.378-1.26 0-2.016.63-2.412 1.152v3.564h3.348V408h-7.236zm15.66.18c-2.052 0-3.132-.792-3.132-3.564v-3.042h-2.538v-1.53h2.538v-3.69h1.926v3.69h3.906v1.53h-3.906v2.97c0 1.494.288 2.016 1.62 2.016.738 0 1.422-.162 1.98-.396v1.548c-.576.252-1.332.468-2.394.468zm5.4 3.186v-1.584c1.728 0 2.574-.54 3.15-1.836l-3.528-7.902h2.16l1.53 3.762.882 2.448.9-2.43 1.548-3.78h2.052l-3.33 7.56c-1.224 2.754-2.394 3.762-5.364 3.762zM95.92 408l3.618-10.296h2.124L105.28 408h-2.142l-.756-2.268h-3.618L98.026 408H95.92zm3.636-4.914l-.342 1.062h2.664l-.342-1.062-.99-3.33-.99 3.33zm8.46 4.914l-1.386-7.956h1.818l.828 6.426 1.134-4.842h2.016l1.17 4.842.81-6.426h1.764L114.784 408h-2.25l-1.134-4.572-1.134 4.572h-2.25zm12.924.18c-1.548 0-2.808-.684-2.808-2.268 0-1.836 1.638-2.322 3.276-2.322.828 0 1.71.144 2.178.288v-.648c0-1.026-.18-1.836-1.656-1.836-1.116 0-1.764.45-1.764 1.386h-1.818c0-2.178 1.782-2.916 3.708-2.916 2.07 0 3.456.882 3.456 3.114v3.294c0 .306.108.414.432.414.216 0 .54-.054.756-.108v1.278a4.128 4.128 0 0 1-1.53.324c-1.008 0-1.296-.522-1.296
```
hdevalence: By saying "the orchestrator", it sounds like there's one function or something that's holding on to the party and dealer states, but they could be on different computers entirely, with the orchestration code handling sending messages over the network.

docs/range-proof-protocol.md
```diff
@@ -1,47 +1,32 @@
...
 
-First, parties are instantiated with their secret values and the range for the range proof. The dealer assigns a unique index \\(j\\) to each party.
+We wanted a foolproof model that would prevent users from performing any step other than the correct next step of the protocol. We treated each party and dealer as a state machine, and modeled each party and dealer state as a distinct type. We used move semantics to ensure that a previous state is consumed as it transitions to the next state. For instance, the [`PartyAwaitingValueChallenge`](../aggregation/party/struct.PartyAwaitingValueChallenge.html) type has a function [`PartyAwaitingValueChallenge::apply_challenge`](../aggregation/party/struct.PartyAwaitingValueChallenge.html#method.apply_challenge), which receives a [`ValueChallenge`](aggregation/messages/struct.ValueChallenge.html) and returns an object of the next type in the protocol ([`PartyAwaitingPolyChallenge`](../aggregation/party/struct.PartyAwaitingPolyChallenge.html)) and a commitment ([`PolyCommitment`](..aggregation/messages/struct.PolyCommitment.html)), and consumes its own state.
```
hdevalence: These three paragraphs are good on their own, but I'm not sure they fit well here, since it jumps from a mathematical description above to the design tradeoffs made, then back to the math. Maybe it would be better to just describe the notation for 1 dealer / m parties here, and put this description in the `aggregation` docs above?

src/lib.rs
```diff
@@ -37,7 +37,9 @@ pub use generators::{Generators, GeneratorsView, PedersenGenerators};
 pub use proof_transcript::ProofTranscript;
 pub use range_proof::RangeProof;
 
-/// API for performing the aggregate-proving multiparty computation protocol.
+/// The `aggregation` module contains API for performing the aggregated multiparty computation protocol.
+///
+#[doc(include = "../docs/aggregation-api.md")]
```
hdevalence: I'm not sure what the order is for combining doc comments and doc includes, and I can't see the sentence in the rendered docs, so it's probably better to have no doc comment and just put the text right into the API md file

src/range_proof/mod.rs
```diff
@@ -228,7 +228,7 @@ impl RangeProof {
 
 /// Compute
 /// \\[
-/// \delta(y,z) = (z - z^{2}) \langle 1, {\mathbf{y}}^{nm} \rangle + z^{3} \langle \mathbf{1}, {\mathbf{2}}^{nm} \rangle
+/// \delta(y,z) = (z - z^{2}) \langle \mathbf{1}, {\mathbf{y}}^{n \cdot m} \rangle - \sum_{j=0}^{m-1} z^{j+3} \cdot \langle \mathbf{1}, {\mathbf{2}}^{n \cdot m} \rangle
```
hdevalence: :+1:

docs/aggregation-api.md
cathieyun: 👍

docs/aggregation-api.md
cathieyun: 👍

docs/range-proof-protocol.md
cathieyun: 👍

PR #81, opened 2018-05-01
Revert nightly version

hdevalence:
per https://github.com/dalek-cryptography/curve25519-dalek/issues/133 the AVX2 backend for `dalek` is broken on recent nightlies.


oleganza:
How come there is no `:stamp:` emoji?

PR #83, opened 2018-05-01
Unify single and aggregate rangeproofs

### LINKED ISSUE #74: Refine parameters / proof contents for aggregated proofs
Two issues that are related enough that they should probably be fixed at the same time:

- the current aggregated RP functions don't take generators as parameters, so they spend effort recompute them every time;
- the current aggregated RP functions don't allow taking existing commitments as inputs (they're generated internally), and keep the commitments as part of the proofs.

### LINKED ISSUE #75: Align aggregated range proofs with single range proofs
Right now we have both a `RangeProof` struct and an `AggregatedProof` struct, and more generally we have two seperate implementations of both proving and verifying.

This is a tracking issue for the task of de-duplicating the proof structs and figuring out to what extent it makes sense to do so for the implementation.

hdevalence:
This removes the `AggregatedProof` struct entirely and moves its validation
logic into the `RangeProof` struct, which now handles both aggregated and
unaggregated proofs.  The former `RangeProof::verify` is now
`RangeProof::verify_single`, a convenience wrapper.

Closes #74 
Closes #75


hdevalence:
One thing I'm not sure about is the `ValueCommitment` message in the MPC protocol. Right now it has `V, A, S` sent from the party to the dealer (and a note that we should change it to only have `A, S`. Should we assume that the dealer has the value commitments `V` up front, in the same way we assume that they're known at verification time?

If we do that, how should we choose names to distinguish between the `V`s (which are Pedersen commitments to the values) and the `ValueCommitment` message (which is a vector pedersen commitment to the bits and to the blinding factors used later)? Calling them both "value commitments" is confusing.

oleganza:
how about `ValueCommitment`, `BitsCommitment` ?

cathieyun:
I don't think that we should assume the dealer has the value commitments `V` up front in the same way we assume they're known at verification time... because at verification time, the `V`s come from a different place (the transaction itself) than the proof, iirc. Where else would the `V` values come from in the proof generation process (they wouldn't have been added to the transaction by then, so why do they need to be treated differently than `A`, `S`?)

I don't have a particular opposition to changing the protocol to assume that the dealer has value commitments `V` up front, I just want to understand why we'd want to do that :)

hdevalence:
They're also used in the proof share verification logic when validating a single proof share. Maybe it's simpler to just keep the `V, A, S` structure?

oleganza:
@cathieyun has a point. I'm don't see a problem with keeping V,A,S together for the purposes of the prover's algorithm.

hdevalence:
Ooops, I didn't see your comment before posting that. I think your reasoning is good. I'll just remove the comment that says "remove this" as part of this PR, then?

cathieyun:
sounds good!

hdevalence:
Another question that I'd like to resolve as part of this change: what should we do with the `SinglePartyAggregator`? Does it make sense to have that at all, or could it just be a method on the `RangeProof` struct, so that there's `create_single` and `create_aggregate` methods on `RangeProof`, and then also the "internal" MPC types that people can use to do multiparty aggregation?

cathieyun:
`create_single` and `create_aggregate` seem fine.

hdevalence:
This is done, and should be followed up by #84 (rearranging the source tree now that things are simplified).

src/range_proof.rs
```diff
@@ -12,209 +11,137 @@ use curve25519_dalek::ristretto::RistrettoPoint;
...
 use util;
 
-use generators::GeneratorsView;
-
 /// The `RangeProof` struct represents a single range proof.
 #[derive(Serialize, Deserialize, Clone, Debug)]
 pub struct RangeProof {
     /// Commitment to the bits of the value
-    A: RistrettoPoint,
+    pub(crate) A: RistrettoPoint,
```
cathieyun: what's the rationale for making this `pub(crate)`?

hdevalence: Originally this was so that the aggregation code could use the internals of the RangeProof struct. It may no longer be necessary and after #84 I'm pretty sure it's not necessary, so that's a good catch. I think it's better to fix it there, though.

hdevalence: Fixed in https://github.com/chain/ristretto-bulletproofs/pull/84/commits/0e7ba65fa0ed9a3706bc1e90b51c9d526e090997 , thanks!

cathieyun: 👍

PR #84, opened 2018-05-01
Rearrange source tree

hdevalence:
After #83, there's no need for the `aggregated_range_proof` tree, but I made this as a seperate PR so that it's easier to review the substantive changes of #83 vs the superficial changes of this one.


hdevalence:
@cathieyun I think that this may move things out from under your notes, but I can do the work of fixing that branch relative to this one.

cathieyun:
No worries - it's not a big change, I can fix that up :)

src/range_proof/mod.rs
```diff
@@ -1,35 +1,74 @@
...
-impl SinglePartyAggregator {
-    /// Create an aggregated rangeproof of multiple values.
+impl RangeProof {
+    /// Create a rangeproof for a given pair of value `v` and
+    /// blinding scalar `v_blinding`.
     ///
-    /// This performs the same proof aggregation MPC protocol, but
-    /// with one party playing all roles.
+    /// XXX add doctests
+    pub fn prove_multiple_single<R: Rng>(
```
oleganza: shouldn't it be named `prove_single`?

src/range_proof/mod.rs
```diff
@@ -70,17 +109,165 @@ impl SinglePartyAggregator {
...
+    ///
+    /// XXX add doctests
+    pub fn verify<R: Rng>(
+        &self,
+        value_commitments: &[RistrettoPoint],
+        gens: GeneratorsView,
+        transcript: &mut ProofTranscript,
+        rng: &mut R,
+        n: usize,
+        m: usize,
```
oleganza: maybe use `m = value_commitments.len()` instead of an explicit parameter?

PR #86, opened 2018-05-03
Batch verification for range proofs

oleganza:
This adds batch-verification to any number of range-proofs. The API ensures that all rangeproofs use the same width `n`, but allows using different aggregation sizes `m`.

Performance depends on Pippenger implementation (not merged yet): https://github.com/dalek-cryptography/curve25519-dalek/pull/129

* Addresses #22. 
* Replaces #27.


oleganza:
Allowing different `n`s might be pretty easy to do as well.

hdevalence:
Instead of moving to a new `verification` module, I think it would be better to keep all of this as part of the rangeproof.

Currently, the `Verification` struct holds a bunch of temporary data. I'm wondering whether we can get away with not holding any temporaries, and have the verification struct act as a view into the proof data, computing things on the fly.

Related to the above is something we delayed thinking about earlier: whether the proof data should have compressed points, so that we don't have to decompress during deser and then recompress to hash during verification. This could impact whether we have to allocate temporaries, but I don't think we made a ticket for this.

oleganza:
> Related to the above is something we delayed thinking about earlier: whether the proof data should have compressed points, so that we don't have to decompress during deser and then recompress to hash during verification. This could impact whether we have to allocate temporaries, but I don't think we made a ticket for this.

@hdevalence Does it depend on upstream support for deserializing compressed points? Do we need to get creative and have `DecompressedRistretto` type that holds both compressed bytes (for fast compression) and the decompressed point? So that we can verify point validity during deserialization and also avoid extra work?

hdevalence:
We could just use compressed points in the struct, and call `.map(|p_bytes| p.decompress())`, with some extra iterator combinators to handle the `None` case.

oleganza:
I rewrote the verification so that API does not expose the helper object with temporaries.

hdevalence:
Just wondering, what's the motivation for making the batch verification API mix all kinds of proofs in a single batch, instead of requiring that the batch contains only the same type of proof?

I'm worried that there's additional complexity and room for mistakes (both in the implementation and in the code that uses it) by mixing proof types, and I'm also not sure what kind of use-case that would be important for.  Even supposing you had a protocol with different kinds of proofs, you could alternately create a batch of proofs of type 1, a batch of proofs of type 2, etc.

cathieyun:
The code looks good, but I am also wondering the same as above ^ - if we don't mix the different proof sizes, it seems like a some of the harder-to-follow logic can be removed (eg figuring out the max generator sizes, the split of `n*m` into `n*m, 1`, padding proofs).

hdevalence:
The point about varying aggregation sizes (`m`) is convincing, and that seems like something we should support in the future.

But as things are now, I think it actually violates the contract of the `Generators` struct, which contains generators for aggregated range proofs of size `(m,n)`. That API isn't designed for varying `m`, and I think that the awkwardness about doing ad-hoc `m*n`s etc., is really a symptom of trying to use the `Generators` in a way that it wasn't designed for.  If we would like to support varying `m`, I think we should redesign the `Generators` API around the idea that `m` can vary, and then update the rest of the code accordingly.

src/range_proof/mod.rs
```diff
@@ -137,17 +137,118 @@ impl RangeProof {
...
+        )
+    }
+
+    /// Verifies multiple range proofs at once.
+    /// If any range proof is invalid, the whole batch is invalid.
+    /// Proofs may use different ranges (`n`) or different number of aggregated commitments (`m`).
+    /// You must provide big enough view into generators (`gens`) that covers
+    /// the biggest proof
+    pub fn verify_batch<'a,'b,I,R,P,V>(
+        proofs: I,
```
hdevalence: I think it would be much simpler just to take a `&[RangeProof]` here.

src/range_proof/mod.rs
```diff
@@ -137,17 +137,118 @@ impl RangeProof {
...
+            rng,
+        )
+    }
+
+    /// Verifies multiple range proofs at once.
+    /// If any range proof is invalid, the whole batch is invalid.
+    /// Proofs may use different ranges (`n`) or different number of aggregated commitments (`m`).
+    /// You must provide big enough view into generators (`gens`) that covers
+    /// the biggest proof
+    pub fn verify_batch<'a,'b,I,R,P,V>(
```
hdevalence: I don't see where `'a, 'b` are used?

src/range_proof/mod.rs
```diff
@@ -137,17 +137,118 @@ impl RangeProof {
...
+            v
+        }).collect::<Vec<_>>();
+
+        if gens.G.len() < nm {
+            return Err(
+                "The generators view does not have enough generators for the largest proof",
+            );
+        }
+
+        // First statement is used without a random factor
```
hdevalence: IMO it is cleaner not to unroll the first loop iteration, and just multiply all statements by a random factor. The code is simpler, the additional cost is minimal (and shrinks as the batch size grows).

src/range_proof/mod.rs
```diff
@@ -184,48 +285,76 @@ impl RangeProof {
...
+                .cloned()
+                .collect(),
         }
     }
 }
 
+/// Represents a deferred computation to verify a single rangeproof.
+/// Multiple instances can be verified more efficient as a batch using
+/// `RangeProof::verify_batch` function.
+struct Verification {
```
hdevalence: In the actual batchverify code above, each of these fields are used only once in the inner loop accumulating the batched verification scalars.

So, instead of having a seperate `Verification` struct that contains owned copies of the verification scalars for a given `RangeProof`, why not just have functions on the `RangeProof` struct that return the iterators for each of the fields of what is currently the `Verification` struct?

goldenMetteyya:
Hi @oleganza what's the status of this PR? One suggestion is to have two apis one for same 'm' and one for a variable one.

PR #87, opened 2018-05-04
Prevent a malicious dealer from retrieving the party's secrets

hdevalence:
@oleganza pointed out that when the dealer's polynomial challenge `x` is zero,
the parties will leak secrets, and suggested this check.

Informal analysis of the protocol flow:

dealer -> party: position

party -> dealer: `ValueCommitment`

    The `ValueCommitment` cannot leak information to the dealer, since
    it's blinded independently of any dealer messages

dealer -> party: `ValueChallenge`

    Contains `y, z` challenges.

party -> dealer: `PolyCommitment`

    Contains `T_1`, `T_2`, which are blinded independently from any
    dealer messages and therefore can't leak information

dealer -> party: `PolyChallenge`

    Contains `x` challenge

party -> dealer: `ProofShare`

    Up till now, we know that the each of the party's messages can't
    reveal info because they're blinded independently of the dealer
    messages.  The paper notes that the blindings for the `l` and `r`
    vectors are chosen such that the prover can reveal `l(x), r(x)`
    for one challenge point `x \in \ZZ_p^{\times}` without revealing
    information, but if `x` is zero, then `\blinding t(x)` is just
    `z^2 \blinding v`, so the dealer could multiply by `z^2` and
    recover the blinding for the original commitment.

    However, if the party checks that `x` is nonzero, then `x \in
    \ZZ_p^{\times}`, the blinding factors are not annihilated, and
    the proof share does not leak information.

In the previous (non-MPC) version of the code, the `x` was computed by the
prover out of the proof transcript, and so it wasn't necessary to check that
`x` is nonzero (which occurs with probability `~ 2^{-252}`): as noted by AGL,
the instructions required to check this are more likely to fail than the check
itself.

Another concern is about replay attacks: a malicious dealer who was able to get
the party to apply the same `PolyChallenge` multiple times could use polynomial
interpolation to recover the coefficients of the `t` polynomial and reveal the
party's secrets.  (I think three points would be sufficient).

However, because our encoding of the protocol flow into affine types ensures
that each `Party` state can be used at most once, we are ensured that any
compilable instantiation of the MPC protocol is invulnerable to replay attacks,
since typechecking the program requires the compiler to prove that states are
not reused.

### DIFF
--- a/src/range_proof/mod.rs
+++ b/src/range_proof/mod.rs
@@ -105,7 +105,8 @@ impl RangeProof {
         let proof_shares: Vec<_> = parties
             .into_iter()
             .map(|p| p.apply_challenge(&poly_challenge))
-            .collect();
+            // Collect the iterator of Results into a Result<Vec>, then unwrap it
+            .collect::<Result<Vec<_>,_>>()?;
 
         dealer.receive_trusted_shares(&proof_shares)
     }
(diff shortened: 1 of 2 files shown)


oleganza:
> as noted by AGL

[citation needed] :-)

hdevalence:
I can't find the citation, but I think it was in some blog post, maybe about Elligator, where something is supposed to be nonzero (but it's the output of a hash function).

PR #90, opened 2018-05-10
Refactor Party into enum, add Clone impl

gterzian:
Warning: annoying bike-shedding proposal incoming... 😄 

While looking around, the 'Party and Dealer state machines' design caught my attention. It's a very effective way of using the language, and it could be even better in the following way:

1. 'more idiomatic' Rust. This is subjective so I won't spend a lot of words on it, let's just say that it feels to me that using an `enum` of `struct` is more idiomatic than having a separate `struct` for each state(while retaining all previous benefits). 
2. A more secure, and flexible, implementation. 

Re 2, while the 'move semantics' of the type system prevents use after move, you are also implicitly relying on `Clone` never being implemented for the various state structs. `Clone` could end-up being derived or implemented in the future, either by accident, carelessness, or outright trickery. You won't get a warning from the compiler if that happens.

Also, while Rust's "orphan rules" prevent a third-party from implementing `Clone` outside of your crate, the rule aims at "trait coherence", not "trait security", it's very complicated, and it might change or have some exceptions in the future(here is a [general overview](http://smallcultfollowing.com/babysteps/blog/2015/01/14/little-orphan-impls/)).

While this rule can probably be relied on for it's core aim of preventing multiple conflicting implementations, it's maybe not best to rely on it to prevent potential malicious parties to implement a trait for one of your types for which the trait has been left unimplemented. 

Hence, you could consider to adopt a more future-proof solution of preemptively implementing `Clone` yourself, ensuring any clone could never be useful for a replay attack, and having the compiler prevent any conflicting implementation. 

Incidentally, this is easily done using the proposed design, highlighting it's flexibility.

I look forward to your thoughts, and would be happy to continue with `Dealer` if you feel it is worthwhile.

### DIFF
--- a/src/range_proof/mod.rs
+++ b/src/range_proof/mod.rs
@@ -90,23 +90,23 @@ impl RangeProof {
         let (parties, value_commitments): (Vec<_>, Vec<_>) = parties
             .into_iter()
             .enumerate()
-            .map(|(j, p)| p.assign_position(j, rng))
+            .map(|(j, p)| p.assign_position(j, rng).unwrap())
             .unzip();
 
         let (dealer, value_challenge) = dealer.receive_value_commitments(&value_commitments)?;
 
         let (parties, poly_commitments): (Vec<_>, Vec<_>) = parties
             .into_iter()
-            .map(|p| p.apply_challenge(&value_challenge, rng))
+            .map(|p| p.apply_value_challenge(&value_challenge, rng).unwrap())
             .unzip();
 
         let (dealer, poly_challenge) = dealer.receive_poly_commitments(&poly_commitments)?;
 
         let proof_shares: Vec<_> = parties
             .into_iter()
-            .map(|p| p.apply_challenge(&poly_challenge))
+            .map(|p| p.apply_poly_challenge(&poly_challenge).unwrap())
             // Collect the iterator of Results into a Result<Vec>, then unwrap it
-            .collect::<Result<Vec<_>,_>>()?;
+            .collect();
 
         dealer.receive_trusted_shares(&proof_shares)
     }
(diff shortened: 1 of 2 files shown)


hdevalence:
Hi! Again, thanks for checking out our code!

We did consider using an enum, but our goal was to have static verification of the protocol flow. As @oleganza pointed out, what we're hoping to enforce with our type system is that the "orchestrating" code does all operations in the correct order, so moving to a dynamic check wouldn't give us what we want.

If I understand your concern about `Clone` and the orphan rules correctly, the concern is that a future change in Rust might cause there to be a `Clone` impl, or that the party state would fall into an exception. But again, what we're guaranteeing is a static property of an implementation, so even if Rust tossed the orphan rules entirely, any currently compilable instantiation of this API would still be correct.

To make this more idiomatic, I think a better path would be to generalize Rust's generator support, so that generators can take inputs. Then we would be able to express what we're doing now using builtin language features, but this isn't currently possible.

We're planning to write a blog post about this soon, but for now I think we're happy with the static guarantees we get.

gterzian:
Thanks you both for your feedback. I'm aware this is the kind of bike-shedding that doesn't change much for your project for now. However, the patterns you adopt now will set the tone for later, and I'de like to argue you're currently not taking the optimal path. 

I would ask you to consider you're  overestimating the value of what you refer to as 'the static check', and giving up lots of upside in terms of flexible and powerful idioms. 

#### The "static" vs "dynamic" distinction you make

> what we're hoping to enforce with our type system is that the "orchestrating" code does all operations in the correct order, so moving to a dynamic check wouldn't give us what we want.

The match and enums will result in runtime checks, which aren't exactly 'dynamic',  they rely on the type system as much as the current implementation. The type system, by way of the enum, is also enforcing the correct order or operations. In what way is this not giving you what you want?

You are, on the other hand, losing a lot of expressiveness and flexibility, the `PartyState::Invalid` variant, even if it is useless, being a point in case. Can you imagine doing that so easily with your current setup?

#### The advantage of enums

Looking at other parts of your codebase, the same pattern of using `struct` when you could be best served with `enum` seems to be repeated.  For example with `Generators` and `PedersenGenerators`, which could be replaced by really just an enum with, for now, only one `Pedersen` variant(and as your project adds support to other commitment schemes, others). 

An example, `Default` could simply have been implemented on `Generators`, with an inner match on the variant to determine what default to return. `Generator::new` could be passed an enum called `InitialGeneratorState::Pedersen` containing the pedersen stuff used to compute the rest. That would give you two sets of enums, one `Generator` and one `InitialGeneratorState`, and enourmous flexibility and ease of understanding for anyone looking at your code.

You'll be missing 'static' checks that you're dealing with a Pedersen generator, and you'll have to match for it across your codebase! That is exactly the pattern that could help your project scale to a large size with lots of 'types' without having your head explode. 

What would it take to currently add support for a different commitment scheme using your current type structure? (this isn't about doing complicate things before they are needed, it's just about following idioms that will naturally allow your code to grow)

#### The security of your API

> even if Rust tossed the orphan rules entirely, any currently compilable instantiation of this API would still be correct.

The API would be correct, yet I was under the impression, from reading [this](https://github.com/chain/ristretto-bulletproofs/commit/57dc2a93e0b125b6f492c385d8f97d61ee54c056#diff-87a21bbd0ba206f9313778a2fd773f81L25) that you also wouldn't want anyone from being able to run any step of the process several times and collect the results. 

Adding `Clone` to your structs(which could happen in various ways), seems to allow such replay. See https://github.com/chain/ristretto-bulletproofs/pull/92

I think there are lots of ways `Clone` could end-up being implemented, one of the most likely ones is someone adding a `#[derive(Clone)]` for an unrelated reason and not realizing this breaks the current security model. Such a mistake will compile just fine, whereas it will not if you implement it first like I suggest(and adding a comment in the implementation explaining what it's for should ensure a commit deleting would not be merged without first ensuring the effect of it).

>  Each party has to trust their own computer to not have incorrect code such as one that somehow implements Clone and then attempts to re-use the state behind your back. 

Yes, yet you already are building a security model using the 'move semantics' to try to prevent replay attacks. Why bother with that? I guess you have at least a responsibility to ensure the integrity of the promise you make as part of your API, and currently this seems to includes disallowing repeating of operations.

cathieyun:
I think it would be informative to look into some research into the session types model that we implemented. There has been a lot of discussion of its security model, and comparisons to other approaches. Here are a few good starting points:
http://munksgaard.me/papers/laumann-munksgaard-larsen.pdf
http://www.di.unito.it/~dezani/papers/sto.pdf

gterzian:
>  It is not a security model, it is a safety model (aka footgun-prevention model)

Thanks for the explanation. One could argue the cloning use-case could fit in that, but let's not go there for now...

@cathieyun Thanks for sharing those articles. I now have a much better understanding of what you are trying to do here...

So essentially you want the type system to 'natively' express that one step can only be taken in response to a specific previous one. 

Obviously, my enum and matching don't qualify, the 'steps' are expressed through the matching, which, while relying on the type system, isn't a 'type' in itself. 

The current implementation also lacks this. The current types don't express in themselves the 'steps' they need to follow, instead that is achieved through the signatures of the methods on them. It's quite close to my proposal which relies on the types chosen in the matching. Both implementations lack the typed concept of 'duality' referred to in those papers. 

I like the "Session-typed Channels" concept in the Danish paper.

It would allow to precisely 'type' not just the state of both `Dealer` and `Party` but also the various steps in their interactions. For example in the case of `DealerAwaitingValueCommitments`, while clearly that type is only able to receive value commitments, the fact that this should happen in response to the parties having been assigned their positions first isn't something the type system currently checks.  

I would be interested in trying to get to an implementation of the party/dealer protocol using such "Session-typed Channels". 

As the most simple first step, I suggest just having a  kind of event-loop like setup that simply uses channels to schedule 'steps' one after the other on the main thread, just as a way to make the conceptual change. I guess this could lend itself naturally to several threads/processes, or over the network, later(which is your goal, right?).

Any takers?

oleganza:
>For example in the case of `DealerAwaitingValueCommitments`, while clearly that type is only able to receive value commitments, the fact that this should happen in response to the parties having been assigned their positions first isn't something the type system currently expresses.

A nitpick: in asynchronous network it's impossible to know (statically or dynamically) whether the other side (party vs dealer) had or had not yet transitioned to "position assigned" state.

What type system perfectly expresses is the sequence of the _local_ (per-actor) state transitions, which is most what one could ask for anyway. Meaning, type `N` can only be produced by a single method on a self-consuming type `N-1`.

PR #99, opened 2018-06-06
Canonical serialization format for RangeProof

oleganza:
This adds a canonical encoding algorithm for a RangeProof in form of `to_bytes`/`from_bytes` API. Serde trait is implemented in terms of these. 

The layout of an `2^n`-bit range proof is defined as:
* 4 compressed Ristretto points `A, S, T_1, T_2`,
* 3 scalars `t_x, t_x_blinding, e_blinding`,
* `2*n` pairs of compressed Ristretto points `L_0, R_0, ..., L_{n-1}, R_{n-1}`,
* two scalars `a` and `b`.

Total size: `32*(9+2*n)` bytes. 

64-bit range proofs are encoded in 672 bytes, 32-bit range proofs are encoded in 608 bytes.

Bonus: the compressed data is cached and reused in verification, so we do not incur unnecessary cost of compressing points that we decompressed just prior to verification.

PS. Serde is intentionally not implemented for IPP as we don't seem to be exposing it directly to users. If we ever need to, we can add ser/de trait implementation like we do with RangeProof.

### DIFF
--- a/src/inner_product_proof.rs
+++ b/src/inner_product_proof.rs
@@ -5,16 +5,17 @@
 use std::borrow::Borrow;
 use std::iter;
 
-use curve25519_dalek::ristretto::RistrettoPoint;
+use curve25519_dalek::ristretto::{CompressedRistretto, RistrettoPoint};
 use curve25519_dalek::scalar::Scalar;
 use curve25519_dalek::traits::VartimeMultiscalarMul;
 
 use proof_transcript::ProofTranscript;
+use util::{read_ristretto, decode_scalar};
 
-#[derive(Serialize, Deserialize, Clone, Debug)]
+#[derive(Clone, Debug)]
 pub struct InnerProductProof {
-    pub(crate) L_vec: Vec<RistrettoPoint>,
-    pub(crate) R_vec: Vec<RistrettoPoint>,
+    pub(crate) L_vec: Vec<CompressedRistretto>,
+    pub(crate) R_vec: Vec<CompressedRistretto>,
     pub(crate) a: Scalar,
     pub(crate) b: Scalar,
 }
@@ -87,8 +88,8 @@ impl InnerProductProof {
                 G_L.iter().chain(H_R.iter()).chain(iter::once(Q)),
             );
 
-            L_vec.push(L);
-            R_vec.push(R);
+            L_vec.push(L.compress());
+            R_vec.push(R.compress());
 
             verifier.commit(L.compress().as_bytes());
             verifier.commit(R.compress().as_bytes());
@@ -130,10 +131,8 @@ impl InnerProductProof {
 
         let mut challenges = Vec::with_capacity(lg_n);
         for (L, R) in self.L_vec.iter().zip(self.R_vec.iter()) {
-            // XXX maybe avoid this compression when proof ser/de is sorted out
-            transcript.commit(L.compress().as_bytes());
-            transcript.commit(R.compress().as_bytes());
-
+            transcript.commit(L.as_bytes());
+            transcript.commit(R.as_bytes());
             challenges.push(transcript.challenge_scalar());
         }
 
@@ -181,7 +180,7 @@ impl InnerProductProof {
         Q: &RistrettoPoint,
         G: &[RistrettoPoint],
         H: &[RistrettoPoint],
-    ) -> Result<(), ()>
+    ) -> Result<(), &'static str>
     where
         I: IntoIterator,
         I::Item: Borrow<Scalar>,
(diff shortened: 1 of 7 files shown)


cathieyun:
👀

src/range_proof/party.rs
```diff
@@ -158,14 +160,14 @@ impl<'a> PartyAwaitingValueChallenge<'a> {
...
-            .pedersen_generators
-            .commit(t_poly.2, t_2_blinding);
+        let T_1 = self.generators.share(self.j).pedersen_generators.commit(
+            t_poly.1,
+            t_1_blinding,
+        );
+        let T_2 = self.generators.share(self.j).pedersen_generators.commit(
+            t_poly.2,
+            t_2_blinding,
+        );
```
hdevalence: I guess these changes are just caused by a `rustfmt` version mismatch?

hdevalence:
These commits change the data for the `RangeProof` and `InnerProductProof` structs to add an extra `Option<Vec<u8>>` to attach an extra copy of the serialized proof data.

Wouldn't it be cleaner to instead change the points held by the proof structs to be `CompressedRistretto` rather than `RistrettoPoint`s, rather than attaching extra data? The proofs never need the uncompressed points except during when checking the verification equations, and the points can just be decompressed inside the verification function.

Then serialization and deserialization can be done with respect to a `&[u8]`/`&mut [u8]` (rather than requiring a `Vec`), and it's just a byte copy.

src/inner_product_proof.rs
```diff
@@ -214,11 +227,84 @@ impl InnerProductProof {
...
+    /// For `n` multiplications the proof size is \\(32 \cdot (2\lg n+2)\\) bytes.
+    pub fn serialized_size(&self) -> usize {
+        (self.L_vec.len() * 2 + 2) * 32
+    }
+
+    /// Serializes the proof into a byte array of \\(2n+2\\) 32-byte elements.
+    /// The layout of the inner product proof is:
+    /// * \\(n\\) compressed Ristretto points \\(L_0, \dots, L_{n-1}\\),
+    /// * \\(n\\) compressed Ristretto points \\(R_0, \dots, R_{n-1}\\),
+    /// * two scalars \\(a, b\\).
```
hdevalence: Does it make sense to interleave the L,R points, since that's the order in which they're generated and fed into the transcript?

oleganza: Yeah, that might be cleaner. Will try it.

oleganza:
> The proofs never need the uncompressed points except during when checking the verification equations, and the points can just be decompressed inside the verification function.

Will try that.

hdevalence:
I guess for serialization it's probably better to return a `Vec`, since making the callers guess at the length seems less than ideal.

oleganza:
We can have `serialized_size()->usize` method for callers to know exact size needed, but since we do a bunch of allocations inside the codebase, it does not help a lot to have clunky `fill-&mut-slice-in-place` API.

hdevalence:
Yeah I agree that that would be pretty inconvenient.

oleganza:
PTAL. Proofs now store L,R,...L,R interleaved and keep `CompressedRistretto` points.

hdevalence:
Rebased.

src/inner_product_proof.rs
```diff
@@ -210,15 +223,75 @@ impl InnerProductProof {
...
         if expect_P == *P {
             Ok(())
         } else {
-            Err(())
+            Err("InnerProductProof is invalid")
+        }
+    }
+
+    /// Returns the size in bytes required to serialize the inner product proof.
+    /// For `n` multiplications the proof size is \\(32 \cdot (2\lg n+2)\\) bytes.
```
hdevalence: I think this should be "for vectors of length n"?

src/range_proof/mod.rs
```diff
@@ -204,27 +221,126 @@ impl RangeProof {
                 .chain(h)
                 .chain(x_sq.iter().cloned())
                 .chain(x_inv_sq.iter().cloned()),
-            iter::once(&self.A)
-                .chain(iter::once(&self.S))
+            iter::once(&self.A.decompress().ok_or(
+                "RangeProof.A is invalid Ristretto point",
```
hdevalence: I think it would be cleaner not to interleave the error handling and the chaining here.

hdevalence:
I fixed up all the nitpicks I had, this looks good to me.

There's an extra allocation when serializing the IPP, which is unfortunate, but we also have to do an extra allocation because of the custom Serde stuff. I don't think it's a big deal.

PR #101, opened 2018-06-11
Arithmetic circuit notes

cathieyun:
Added the notes on how circuit math works - PTAL

Next steps - organize the notes better (table of contents or break into multiple files)
(remember to change references & copy duplicate footnotes)

### DIFF
--- a/docs/notes.md
+++ b/docs/notes.md
@@ -2,11 +2,14 @@ This module contains notes on how and why Bulletproofs work.
 
 The documentation is laid out roughly as follows.  General notes on
 the range proof and inner-product proofs are here.  The description of
-each protocol is contained in the respective `range_proof` and
-`inner_product_proof` modules.  Finally, structs from those modules
+each protocol is contained in the respective [`range_proof`][rp_notes] and
+[`inner_product_proof`][ipp_notes] modules.  Finally, structs from those modules
 are publicly re-exported from the crate root, so that the external
 documentation describes how to use the API, while the internal
 documentation describes how it works.
+
+[rp_notes]: https://doc-internal.dalek.rs/ristretto_bulletproofs/range_proof/index.html
+[ipp_notes]: https://doc-internal.dalek.rs/ristretto_bulletproofs/inner_product_proof/index.html
 
 Notation
 ========
@@ -612,7 +615,7 @@ The aggregation protocol is a multi-party computation protocol, involving \\(m\\
 
 The Bulletproofs paper outlines two versions of multi-party computation aggregation. In the first approach, the inner-product proof is performed by the dealer, which requires sending the vectors used for the inner-product to the dealer. In the second approach, the inner-product proof is performed using multi-party computation, which sends less data but requires one round for each iteration of the inner-product protocol. We chose to implement the first approach because it requires fewer round trips between parties, which outweighed the slight message size savings of the second approach. 
 
-For more information on how the aggregation protocol works and is implemented, see the [protocol notes](../aggregated_range_proof/index.html). 
+For more information on how the aggregation protocol works and is implemented, see the [protocol notes](../range_proof/index.html). 
 
 The aggregated range proof has the same form as the individual range proof, in that the provers (the parties) still perform the same calculations to prove that \\(t(x) = \langle \mathbf{l}(x), \mathbf{r}(x) \rangle \\) and that \\(t_0, \mathbf{l}(x), \mathbf{r}(x)\\) are correct. The difference is that the challenge values are obtained from the dealer, which generates them by combining commitments from all the parties, and that the calculations of different parties are seperated by different powers of the challenge scalars \\(y\\) and \\(z\\).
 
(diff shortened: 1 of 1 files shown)


docs/notes.md
```diff
@@ -878,5 +881,438 @@ With these observations, we can simplify the combined \\(m\\)-party statement ab
 \end{aligned}
 \\]
 
+Arithmetic Circuit Proofs
+=========================
+
+The goal of an *arithmetic circuit proof* is for a prover to convince a verifier that a particular set of values \\(v\\) satisfy the constraints represented by the arithmetic circuit, without revealing any additional information about the values \\(v\\).
```
hdevalence: I think it might be good here to briefly say what an arithmetic circuit is, before saying what we're trying to prove about one? A few paragraphs down there's mention of "one type of gate", but there's not a prior definition of a circuit as a DAG of addition and multiplication gates.

docs/notes.md
```diff
@@ -878,5 +881,438 @@ With these observations, we can simplify the combined \\(m\\)-party statement ab
...
+\\[
+\begin{aligned}
+    \beta         &\xrightarrow{} \tilde{o} \\\\
+\end{aligned}
+\\]
+
+Proving statements about arithmetic circuits gates
+--------------------------------------------------
+
+One type of gate in an arithmetic circuit is a multiplication gate of fan-in 2, which takes two input values and multiplies them to get an output value. If for all of the multiplication gates in a circuit, \\(\textbf{a}\_L\\) is the vector of left inputs, \\(\textbf{a}\_R\\) is the vector of right inputs, and \\(\textbf{a}\_O\\) is the vector of outputs, then the following equation will represent the relationship between the inputs and outputs of all the multiplication gates in the circuit:
```
hdevalence: So far we just have a vector of values `v`, which are the commitments to the input wires, but now there are three vectors corresponding to the left/right inputs and outputs of all of the gates.  I think it would be good to have described the connection between the circuit DAG (with input wires only at the beginning) and the R1CS setup (where the gates are all pulled apart and then re-connected by consistency equations) before defining what the left/right/output vectors are.

docs/notes.md
```diff
@@ -878,5 +881,438 @@ With these observations, we can simplify the combined \\(m\\)-party statement ab
...
+
+\\[
+\textbf{a}\_L \circ \textbf{a}\_R = \textbf{a}\_O
+\\]
+
+Other types of gates in arithmetic circuits are constant multiplication gates and addition gates. A constant multiplication gate can be represented by a constant multiplied by an input, and an addition gate can be represented by the addition of two or more inputs. All of the constant multiplication gates and addition gates in a circuit can be represented as a collection of linear constraints, as expressed in the following equation:
```
hdevalence: I think that if (parts of this) were moved up to the beginning, you would be able to define an arithmetic circuit in general, then refine the definition to differentiate mul-by-const from mul, and then describe how to get from the DAG repr to the R1CS repr

docs/notes.md
```diff
@@ -878,5 +881,438 @@ With these observations, we can simplify the combined \\(m\\)-party statement ab
...
+\textbf{W}\_O \cdot \textbf{a}\_O =
+\textbf{W}\_V \cdot \textbf{v} +
+\textbf{c}
+\\]
+
+Combining statements using challenge variables
+----------------------------------------------
+
+We can rewrite the statement about multiplication gates into an inner product equation, using the challenge variable \\(y\\). We can do this for a random challenge \\(y\\) because \\({\mathbf{b}} = {\mathbf{0}}\\) if and only
+if[^2] \\({\langle {\mathbf{b}}, {\mathbf{y}}^{n} \rangle} = 0\\).
```
hdevalence: Hmm, it looks like rustdoc doesn't do well with non-unique footnote numbers: clicking this takes you down to the bottom, but hitting the return link sends you back to the beginning of the RP notes.

docs/notes.md
```diff
@@ -878,5 +881,438 @@ With these observations, we can simplify the combined \\(m\\)-party statement ab
...
+
+Combining statements using challenge variables
+----------------------------------------------
+
+We can rewrite the statement about multiplication gates into an inner product equation, using the challenge variable \\(y\\). We can do this for a random challenge \\(y\\) because \\({\mathbf{b}} = {\mathbf{0}}\\) if and only
+if[^2] \\({\langle {\mathbf{b}}, {\mathbf{y}}^{n} \rangle} = 0\\). 
+
+\\[
+\langle \textbf{a}\_L \circ \textbf{a}\_R - \textbf{a}\_O ,
+\textbf{y}^n \rangle = 0
```
hdevalence: I think this formula is kind of hanging out and could be linked up with the previous sentence a little better

docs/notes.md
cathieyun: I fixed it up by having the footnotes be unique. It's currently awkward because there are two sets of footnotes that are the exact same, but when we split the notes into separate files (one for range proof, one for circuit proof, one for inner product proof) it make more sense.

cathieyun:
I updated the circuit -> constraint descriptions, it's much clearer now (thanks for the pointers!) PTAL again?

cathieyun:
PTAL?

docs/notes.md
```diff
@@ -878,5 +881,459 @@ With these observations, we can simplify the combined \\(m\\)-party statement ab
...
+\langle z \textbf{z}^Q,
+\textbf{c} + \textbf{W}\_V \cdot \textbf{v} \rangle + \delta(y, z) =
+\langle \textbf{a}\_L + \textbf{y}^{-n} \circ (z \textbf{z}^Q \cdot \textbf{W}\_R), 
+\textbf{y}^n \circ \textbf{a}\_R \rangle + 
+\langle \textbf{a}\_L,
+z \textbf{z}^Q \cdot \textbf{W}\_L \rangle +
+\langle \textbf{a}\_O, 
+-\textbf{y}^n + z \textbf{z}^Q \cdot \textbf{W}\_O \rangle + 
+\langle \textbf{y}^{-n} \circ (z \textbf{z}^Q \cdot \textbf{W}\_R), z \textbf{z}^Q \cdot \textbf{W}\_L \rangle
+\\]
```
hdevalence: The RHS of this equation is pretty long and it might be better to break it over multiple lines.

PR #102, opened 2018-06-13
Arithmetic circuit proof code

cathieyun:
Corresponds with: https://github.com/chain/ristretto-bulletproofs/issues/97

Circuit proof code:
- generates a circuit proof using IPP optimization
- verifies a circuit proof 
- contains tests for basic multiplication, addition, and shuffle circuits

Next steps:
- figure out how to bypass matrix representation of circuits (R1CS directly to bulletproofs)
- figure out how to do higher-level circuit representations (bellman?)
- benchmarking for more complex circuits

PTAL!


src/circuit_proof/mod.rs
```diff
@@ -0,0 +1,713 @@
...
+    pub T_1: RistrettoPoint,
+    pub T_3: RistrettoPoint,
+    pub T_4: RistrettoPoint,
+    pub T_5: RistrettoPoint,
+    pub T_6: RistrettoPoint,
+    pub t_x: Scalar,
+    pub t_x_blinding: Scalar,
+    pub e_blinding: Scalar,
+    pub l_vec: Vec<Scalar>,
+    pub r_vec: Vec<Scalar>,
```
hdevalence: Are `l_vec` and `r_vec` supposed to be included in the proof, if we're doing the inner product argument?

cathieyun: oh nope - I forgot to remove those after adding the ipp, I'll fix that

src/circuit_proof/mod.rs
```diff
@@ -0,0 +1,713 @@
...
+        q: usize,
+
+        W_L: Vec<Vec<Scalar>>, // Q vectors, of length n each
+        W_R: Vec<Vec<Scalar>>,
+        W_O: Vec<Vec<Scalar>>,
+        W_V: Vec<Vec<Scalar>>, // Q vectors, of length m each
+        a_L: Vec<Scalar>,
+        a_R: Vec<Scalar>,
+        a_O: Vec<Scalar>,
+        v_blinding: Vec<Scalar>,
```
hdevalence: All of `n, m, q, W_L, W_R, W_O, W_V` are circuit parameters, while the `a_L, a_R, a_O` are derived from the witness, right?

Does it make sense to split out the non-witness-related parameters into a single `Circuit` object, instead of having the callers pass all of them manually?

src/circuit_proof/mod.rs
```diff
@@ -0,0 +1,713 @@
...
+        let x = transcript.challenge_scalar(); 
+
+        transcript.commit(self.t_x.as_bytes());
+        transcript.commit(self.t_x_blinding.as_bytes());
+        transcript.commit(self.e_blinding.as_bytes());
+        let w = transcript.challenge_scalar();
+
+        let r = Scalar::random(rng);
+
+        // Calculte points that represent the matrices
```
hdevalence: typo

src/circuit_proof/mod.rs
cathieyun: that is correct - yes, that makes it a lot cleaner! Will do

src/lib.rs
```diff
@@ -28,11 +28,13 @@ mod util;
 
 #[doc(include = "../docs/notes.md")]
 mod notes {}
+mod circuit_proof;
 mod generators;
 mod inner_product_proof;
 mod proof_transcript;
 mod range_proof;
 
+pub use circuit_proof::CircuitProof;
```
hdevalence: Maybe we should leave the `circuit_proof` code unexported until we figure out what kind of API we want people to use for circuit proving?

My intuition so far is that we probably don't want to expose an API like the one we have now, so if we keep the modules unexported we can work towards having a releasable library for range proofs, without having to commit to a circuit proving API before it's done.

src/inner_product_proof.rs
```diff
@@ -51,6 +51,15 @@ impl InnerProductProof {
...
         let mut n = G.len();
 
+        if n == 0 {
+            return InnerProductProof {
+                L_vec: vec![],
+                R_vec: vec![],
+                a: Scalar::zero(),
+                b: Scalar::zero(),
+            };
+        }
```
hdevalence: What is the meaning of an `InnerProductProof` about empty vectors? Is this just to cover the case where a circuit has no multiplication gates?

src/circuit_proof/mod.rs
```diff
@@ -0,0 +1,781 @@
...
+    pub T_6: RistrettoPoint,
+    pub t_x: Scalar,
+    pub t_x_blinding: Scalar,
+    pub e_blinding: Scalar,
+    pub ipp_proof: InnerProductProof,
+}
+
+impl CircuitProof {
+    pub fn prove<R: Rng + CryptoRng>(
+        gen: &Generators,
```
hdevalence: The existing `Generators` struct is designed to hold generators for an `m`-aggregated `n`-bit rangeproof, and requires that `m` and `n` both be powers of two.

But here, the `Generators` struct isn't being used that way, it's being used to hold generators for a single circuit with `n` multiplication gates, and `n` isn't required to be a power of two.

src/circuit_proof/mod.rs
```diff
@@ -0,0 +1,781 @@
...
+                .chain(gen.H.iter())
+                .chain(self.ipp_proof.L_vec.iter())
+                .chain(self.ipp_proof.R_vec.iter())
+                .chain(V.iter())
+                .chain(iter::once(&self.T_1))
+                .chain(iter::once(&self.T_3))
+                .chain(iter::once(&self.T_4))
+                .chain(iter::once(&self.T_5))
+                .chain(iter::once(&self.T_6)),
+        );
```
hdevalence: I think this equation could be simplified and clarified by breaking its components up into pieces and then joining the pieces (like `t_scalars, T_points`, etc).

src/circuit_proof/mod.rs
```diff
@@ -0,0 +1,781 @@
...
+    }
+}
+
+// Computes z * z^Q * W, where W is a qxn matrix and z is a scalar.
+// Input: Qxn matrix of scalars and scalar z
+// Output: length n vector of Scalars
+pub fn matrix_flatten(
+    W: Vec<Vec<Scalar>>,
+    z: Scalar,
+    output_dim: usize,
```
hdevalence: What's the role of the `output_dim` parameter? Since the matrix has a known size, why not use that size instead of making the caller pass a parameter in and check consistency?

src/circuit_proof/mod.rs
```diff
@@ -0,0 +1,781 @@
...
+        let Q = w * gen.pedersen_generators.B;
+
+        let ipp_proof = InnerProductProof::create(
+            transcript,
+            &Q,
+            util::exp_iter(y.invert()),
+            gen.G.to_vec(),
+            gen.H.to_vec(),
+            l_vec.clone(),
+            r_vec.clone(),
```
hdevalence: The existing `InnerProductProof` code is expecting that the vector lengths are a power of two, but this isn't the case any more in the circuit code.  How do we want to handle non-power-of-two inner products?

src/circuit_proof/mod.rs
```diff
@@ -0,0 +1,781 @@
...
+#[derive(Clone, Debug)]
+pub struct Circuit {
+    n: usize,
+    m: usize,
+    q: usize,
+    c: Vec<Scalar>,        // vector of constants, length q
+    W_L: Vec<Vec<Scalar>>, // q vectors, of length n each
+    W_R: Vec<Vec<Scalar>>,
+    W_O: Vec<Vec<Scalar>>,
+    W_V: Vec<Vec<Scalar>>, // q vectors, of length m each
```
hdevalence: I think it would be good to put a note that we should swap out this representation later when we have R1CS support, since for large circuits we almost certainly don't want to actually keep all of these matrices in memory.

src/circuit_proof/mod.rs
cathieyun: The reason we need the `output_dim` parameter is if the matrix is size `qxn` where `q=0` and `n>0` (such as in the `mul_circuit_2` test case). In this case, you could not figure out what the output vector's size should be (since you can't figure out `n` by looking at the matrix), but the output of the `matrix_flatten` function should still be a length `n` vector. 

It's possible to do a check higher up for if `n=0`, before calling `matrix_flatten`, but I thought it would be cleaner to pass in `n` as `output_dim` and have the `matrix_flatten` function take care of those checks internally.

src/circuit_proof/mod.rs
cathieyun: I joined the `T_scalars` and `T_points`, that's a good idea.
I think any further simplification / flattening would make it harder to follow along with the math though, since the current version is a pretty direct reflection of the math in the paper/notes.

src/circuit_proof/mod.rs
hdevalence: Hmm, you're right that the `Generators` struct doesn't check that `n, m` are powers of two. I think that's a bug.

My concern is that the `Generators` struct is designed to be used for `m`-aggregated `n`-bit rangeproofs, and here we're not doing that.  So, I think that we should either use a different type for circuit generators (which requires thinking about how we want to relate the circuit generators to the circuit description), or change the `Generators` struct so that it's not just for rangeproofs (which requires checking that this doesn't have any other effects on the rest of the code).

src/inner_product_proof.rs
hdevalence: Great. We should probably double-check the behaviour of the function and write a note about what inputs are allowed (what length of vectors, whether they have to be powers of two, etc).

hdevalence: Maybe we can come up with a reason why this function is OK even when the inputs aren't power-of-two lengths.

hdevalence:
We can defer dealing with the generators and inner-product details for later TODO chunks on the `circuit` branch

PR #103, opened 2018-07-02
Use {lo,hi} instead of {L,R} where appropriate in notes module.

jasondavies:
Other docs seem to use lo and hi, but at least this is self-consistent.  Might be worth picking one notation throughout all modules?

### DIFF
--- a/docs/notes.md
+++ b/docs/notes.md
@@ -529,10 +529,10 @@ and compress the vectors by adding the left and the right halves
 separated by the variable \\(u\_k\\):
 \\[
 \begin{aligned}
-  {\mathbf{a}}^{(k-1)} &= {\mathbf{a}}\_L \cdot u\_k        + u^{-1}\_k \cdot {\mathbf{a}}\_R \\\\
-  {\mathbf{b}}^{(k-1)} &= {\mathbf{b}}\_L \cdot u^{-1}\_k   + u\_k \cdot {\mathbf{b}}\_R \\\\
-  {\mathbf{G}}^{(k-1)} &= {\mathbf{G}}\_L \cdot u^{-1}\_k   + u\_k \cdot {\mathbf{G}}\_R \\\\
-  {\mathbf{H}}^{(k-1)} &= {\mathbf{H}}\_L \cdot u\_k        + u^{-1}\_k \cdot {\mathbf{H}}\_R 
+  {\mathbf{a}}^{(k-1)} &= {\mathbf{a}}\_{\operatorname{lo}} \cdot u\_k        + u^{-1}\_k \cdot {\mathbf{a}}\_{\operatorname{hi}} \\\\
+  {\mathbf{b}}^{(k-1)} &= {\mathbf{b}}\_{\operatorname{lo}} \cdot u^{-1}\_k   + u\_k \cdot {\mathbf{b}}\_{\operatorname{hi}} \\\\
+  {\mathbf{G}}^{(k-1)} &= {\mathbf{G}}\_{\operatorname{lo}} \cdot u^{-1}\_k   + u\_k \cdot {\mathbf{G}}\_{\operatorname{hi}} \\\\
+  {\mathbf{H}}^{(k-1)} &= {\mathbf{H}}\_{\operatorname{lo}} \cdot u\_k        + u^{-1}\_k \cdot {\mathbf{H}}\_{\operatorname{hi}}
 \end{aligned}
 \\]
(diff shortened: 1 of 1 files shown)


hdevalence:
Hi, thanks for the PR!

Originally we were using `_L` and `_R` for the "left" and "right" halves of the vectors, because we didn't like the slice notation that the paper uses. This turned out to not be the best notation, since the distinction between the left and right sides of the inner product is important, and it's confusing to have two different senses of "left" and "right". So we changed to `_{lo}` and `_{hi}` for the low and high halves of the vectors.

All this is to say that we'd like to be consistent, but rather than changing `lo` and `hi` to `L` and `R`, it would be better to change `L` and `R` to `lo` and `hi`.

jasondavies:
Ah, that makes sense!  If you'd like, I can update this PR to use `lo` and `hi` instead of `L` and `R` in the notes module.  Or you may prefer to do it yourself?  Either way, happy to help out.

hdevalence:
If you'd like to make the changes yourself, feel free!

docs/notes.md
```diff
@@ -66,13 +66,13 @@ instead of \\(\operatorname{Com}(v, {\widetilde{v}})\\) for brevity.
 We also make use of *vector Pedersen commitments*, which we define for
 pairs of vectors as \\[
 \begin{aligned}
-    \operatorname{Com}({\mathbf{a}}\_{L}, {\mathbf{a}}\_{R}) 
- &= \operatorname{Com}({\mathbf{a}}\_{L}, {\mathbf{a}}\_{R}, {\widetilde{a}})
-  = {\langle {\mathbf{a}}\_{L}, {\mathbf{G}} \rangle} + {\langle {\mathbf{a}}\_{R}, {\mathbf{H}} \rangle} + {\widetilde{a}} {\widetilde{B}},\end{aligned}
+    \operatorname{Com}({\mathbf{a}}\_{\operatorname{lo}}, {\mathbf{a}}\_{\operatorname{hi}})
```
hdevalence: These `a_L` and `a_R` are different vectors (left and right bit commitments), not the low and high halves of a single vector...

jasondavies: Ah, oops. Will fix...

docs/notes.md
```diff
@@ -315,7 +315,7 @@ we can express the coefficients of \\(t(x)\\) using Karatsuba’s method:
 \\]
 Since \\[
 \begin{aligned}
-  t\_{0} &= {\langle {\mathbf{a}}\_{L} - z {\mathbf{1}}, {\mathbf{y}}^{n} \circ ({\mathbf{a}}\_{R} + z {\mathbf{1}}) + z^{2} 2^{n} \rangle},\end{aligned}
+  t\_{0} &= {\langle {\mathbf{a}}\_{\operatorname{lo}} - z {\mathbf{1}}, {\mathbf{y}}^{n} \circ ({\mathbf{a}}\_{\operatorname{hi}} + z {\mathbf{1}}) + z^{2} 2^{n} \rangle},\end{aligned}
```
oleganza: Here `a_L` and `a_R` are distinct vectors, not two halves of a single vector.

oleganza:
Apart from the `L/R<->lo/hi` confusion on line 318, the rest looks good.

docs/notes.md
oleganza:
```suggestion
  t\_{0} &= {\langle {\mathbf{a}}\_{L} - z {\mathbf{1}}, {\mathbf{y}}^{n} \circ ({\mathbf{a}}\_{R} + z {\mathbf{1}}) + z^{2} 2^{n} \rangle},\end{aligned}
```

jasondavies:
Great! I've reverted the change to line 318 now.

oleganza:
@jasondavies hey, could you pls rebase this on the latest `main` branch (505556b218d6f78f4767c38654ca5e5904b08b9e), so we can merge? I just checked, there are no conflicts, so it should be a quick thing to do.

```
git checkout main
git pull git@github.com:dalek-cryptography/bulletproofs.git main
git checkout notes
git rebase main
git push -f git@github.com:plutomonkey/ristretto-bulletproofs.git notes
```

jasondavies:
Done!

hdevalence:
trying to restart travis by closing and reopening this PR, since the build has gone missing

hdevalence:
(couldn't get Travis to remember the PR's existence, but it only changes docs, so I force merged it)

PR #106, opened 2018-07-10
R1CS circuit frontend

cathieyun:
Part of r1cs circuit support plan: https://github.com/dalek-cryptography/bulletproofs/issues/118

This PR creates:
- r1cs format
- r1cs -> BP naive transformation
- r1cs circuits (e.g. shuffle and merge/split), and tests that they generate & verify proofs correctly

### DIFF
--- a/src/circuit_proof/r1cs.rs
+++ b/src/circuit_proof/r1cs.rs
@@ -1,0 +1,17 @@
+use curve25519_dalek::scalar::Scalar;
+use curve25519_dalek::ristretto::RistrettoPoint;
+
+/// Represents a variable in our constraint system.
+#[derive(Copy, Clone, Debug)]
+pub struct Variable(Index);
+
+/// Represents the index of either a public or private variable.
+#[derive(Copy, Clone, PartialEq, Debug)]
+pub enum Index {
+    Public(usize),
+    Private(usize)
+}
+
+/// This represents a linear combination of some variables, with scalar coefficients.
+#[derive(Clone)]
+pub struct LinearCombination(Vec<Variable, Scalar>);
--- a/src/circuit_proof/r1cs.rs
+++ b/src/circuit_proof/r1cs.rs
@@ -1,17 +1,108 @@
 use curve25519_dalek::scalar::Scalar;
 use curve25519_dalek::ristretto::RistrettoPoint;
 
-/// Represents a variable in our constraint system.
-#[derive(Copy, Clone, Debug)]
-pub struct Variable(Index);
+use circuit_proof::{Circuit, CircuitInput};
 
-/// Represents the index of either a public or private variable.
-#[derive(Copy, Clone, PartialEq, Debug)]
-pub enum Index {
-    Public(usize),
-    Private(usize)
+// This is a stripped-down version of the Bellman r1cs representation, for the purposes of
+// learning / understanding. The goal is to write this as a BulletproofsConstraintSystem that 
+// implements the Bellman ConstraintSystem trait, so we can use that code/logic.
+// (That would require the bellman code to be decoupled from the underlying pairings.)
+
+/// Represents a variable in our constraint system, where the value represents the index.
+pub struct Variable(usize);
+
+/// Represents a linear combination of some variables multiplied with their scalar coefficients, 
+/// plus a scalar. E.g. LC = variable[0]*scalar[0] + variable[1]*scalar[1] + scalar
+pub struct LinearCombination {
+	variables: Vec<(Variable, Scalar)>,
+	constant: Scalar, 
 }
 
-/// This represents a linear combination of some variables, with scalar coefficients.
-#[derive(Clone)]
-pub struct LinearCombination(Vec<Variable, Scalar>);
+impl LinearCombination {
+	// TODO: make constructor with iterators
+	// see FromIterator trait - [(a1, v1), (a2, v2)].iter().collect() (pass in the iterator, collect to get LC)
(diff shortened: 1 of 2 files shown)


src/circuit_proof/r1cs.rs
```diff
@@ -0,0 +1,649 @@
...
+        create_and_verify_helper(circuit, prover_input, verifier_input)
+    }
+
+    #[test]
+    // Test that a 2 in 2 out shuffle circuit behaves as expected
+    fn shuffle_circuit() {
+        let three = Scalar::from_u64(3);
+        let seven = Scalar::from_u64(7);
+        assert!(shuffle_circuit_helper(three, seven, seven, three).is_ok());
+        assert!(shuffle_circuit_helper(three, seven, seven, three).is_ok());
```
oleganza: These two lines look identical. Did you mean to permute the order of the last two arguments?

cathieyun: whoops - yes, I'll fix that. good catch!

oleganza:
Maybe it's out of scope for this PR, but I have hard time understanding the boundary between the prover and the verifier. E.g. how does the verifier construct the shuffle circuit on its own?

src/circuit_proof/r1cs.rs
```diff
@@ -0,0 +1,667 @@
...
+/// plus a scalar. E.g. LC = variable[0]*scalar[0] + variable[1]*scalar[1] + scalar
+pub struct LinearCombination {
+    variables: Vec<(Variable, Scalar)>,
+    constant: Scalar,
+}
+
+impl LinearCombination {
+    // TODO: make constructor with iterators
+    // see FromIterator trait - [(a1, v1), (a2, v2)].iter().collect() (pass in the iterator, collect to get LC)
+    pub fn construct(variables: Vec<(Variable, Scalar)>, constant: Scalar) -> Self {
```
hdevalence: I think this should probably be called `new` to be consistent with other Rust constructors

cathieyun:
Re @oleganza's question - currently, there is no good way for the verifier to construct a circuit on its own (this PR was just the MVP for r1cs.) 

But in a future PR, I think it would be a great idea to have (directly copying from a chat with oleg:) a generic circuit API/struct parameterized by T which is either a pair of a secret and its commitment, or just a commitment. Then, compiler will generate two implementations: one for prover, one for verifier; but the source code will have the same function allocating variables, generating challenges and hooking up linear constraints.

src/circuit_proof/r1cs.rs
```diff
@@ -0,0 +1,667 @@
...
+    }
+    // Allocate a variable and do value assignment at the same time
+    pub fn alloc_variable(&mut self, val: Scalar) -> Variable {
+        self.var_assignment.push(val);
+        Variable(self.var_assignment.len() - 1)
+    }
+
+    // Push one set of linear constraints (a, b, c) to the constraint system.
+    // Pushing a, b, c together prevents mismatched constraints.
+    pub fn push_lc(
```
hdevalence: Could this be called `constrain` (or something else that makes the verb more specific than `push`?)

src/circuit_proof/r1cs.rs
```diff
@@ -0,0 +1,667 @@
...
+
+use circuit_proof::{Circuit, ProverInput, VerifierInput};
+
+// This is a stripped-down version of the Bellman r1cs representation, for the purposes of
+// learning / understanding. The goal is to write this as a BulletproofsConstraintSystem that
+// implements the Bellman ConstraintSystem trait, so we can use that code/logic.
+// (That would require the bellman code to be decoupled from the underlying pairings.)
+
+/// Represents a variable in our constraint system, where the value represents the index.
+#[derive(Clone, Debug)]
```
hdevalence: This could also derive `Copy, PartialEq, Eq`

hdevalence:
This looks great so far! The inline comments I put are mainly about naming.

One design suggestion I have is about the `Variable` struct. The index is kept private to the struct, but then exposed by the `get_index` function.  This doesn't seem quite right, because the index corresponds to internal data that's owned by the `ConstraintSystem`, and having the index isn't useful without that data. I think that if the code were restructured to eliminate `get_index`, it would force other changes that make the design a bit cleaner.

For instance, `get_index` is used by the `LinearCombination::check_variables` method. But since the inner part of the `Variable` tuple is private, `Variable`s can only be created by `ConstraintSystem::alloc_variable`, and that function always returns `Variable`s which are valid with respect to that `ConstraintSystem` instance.  So, as long as there's a single `ConstraintSystem` instance, we already know that all possible `Variables` are always valid, and the `LinearCombination::check_variables` function isn't needed.  (If there are multiple `ConstraintSystem` instances, a `Variable` allocated by one wouldn't be valid with respect to another `ConstraintSystem`, but the existing `check_variables` function doesn't catch this either).

`get_index` also used by `LinearCombination::eval`. But evaluation is done using the variable assignments, which belong to the `ConstraintSystem`, so the only part of the code that can call `eval` anyways is the `ConstraintSystem` (which is also the only part of the code that needs to evaluate the LCs). So `eval` could instead be a private method of the `ConstraintSystem` struct.

cathieyun:
@hdevalence agreed, having `get_index()` on the `Variable` struct is awkward. I changed that - now, the only places that access the index of the variable are:
- `create_circuit` function on `ConstraintSystem` which needs the variable index to create the `W_V` matrix
- `eval_lc` function on `ConstraintSystem` (instead of as a function on `LinearConstraint`) which needs the variable index to get the scalar for the variable.

So, the variable index isn't being accessed by anything outside of the `ConstraintSystem` struct now.
PTAL and let me know if that addresses the issue!

PR #109, opened 2018-07-11
Circuit work tracker [WIP]

cathieyun:
This PR tracks work on the `circuit` branch, which is the central branch for circuit/R1CS-related work.


hdevalence:
Changing the base to `develop` since we're not working in two places now

cathieyun:
👍

PR #111, opened 2018-07-12
Merge main into circuit

hdevalence:
This pulls in the changes to use compressed points in the rangeproofs and inner product proofs.

The current circuit proof code stores the points in the circuit proof in uncompressed format, which is now inconsistent with the range proof code.  Maybe this PR should also change the circuit proofs to use compressed points?


hdevalence:
Oh, okay, I was going to add that to this PR. 

I don't think it's super important, but it seems like when merging changes from `main` into `circuit`, we should re-structure the `circuit` code as needed as part of the merge PR, so that the changes are more atomic.

cathieyun:
ok that sounds good for future changes!

Issue #107, opened 2018-07-11
Enforce that circuit `n` parameter is zero or a power of two


cathieyun:
Currently, the inner product proof only works for vectors with lengths that are powers of two.
This was fine for the range proof code, where this was a valid assumption.
However, for circuits, the vectors will not always be powers of 2. 
After some discussion, we decided that the best solution in that case is to pad the circuits such that `circuit.n` is a power of 2, so that the vectors being input to the IPP are also lengths of powers of 2.

Initial discussion here: https://github.com/dalek-cryptography/bulletproofs/pull/102

TODO: double check that empty IPP outputs are okay

hdevalence:
Since the IPP uses variable-time operations, padding with zeros shouldn't cost too much (it only affects the first loop iteration anyways). I think it would be good for the IPP code to check that its inputs are powers of two, and have the callees do the padding if necessary.

cathieyun:
Work here: https://github.com/dalek-cryptography/bulletproofs/pull/115

cathieyun:
Updated thoughts:
I think that it does not make sense to do the circuit padding within the circuit prover module - it is better for the `circuit` module to only accept circuits where `circuit.n` is already a power of two. Then, it is up to whoever is creating a circuit (e.g. by hand, or from `r1cs` format) to make sure that the circuit's `circuit.n` is a power of two.

This makes more sense than doing the check and conversion within the `circuit` module itself, because doing the latter causes a strange mismatch of expectations - circuits can have `circuit.n` as not a power of two, but the circuit creator still needs to pass in a `Generator` that has a length that is a power of two. So, the user of the API still needs to convert _some_ of their inputs (but not all!) to be a power of two. This is confusing - they should either not have to worry about power-of-two properties at all, or have them strictly enforced for all inputs.

Also, the `r1cs` to bulletproofs converter is already creating the matrices for the circuit, so it is very natural to do the padding during the creation process.

So, the proposed plan is to enforce that any circuits being created in the `circuit` module have to have `circuit.n` and `Generators.len()` be a power of two - all other inputs will be rejected. Then, in the `r1cs` module, the circuit creator will pad circuits until they are the right size.

hdevalence:
I think that going forward we're going to want to subsume the `circuit` module entirely into the `r1cs` module -- we should pass directly from the R1CS to the proof, without any intermediate matrices, and we should do the padding as necessary just before creating the proof.

[...]

PR #115, opened 2018-07-16
Pad circuits so `n` is always a power of 2 [WIP]

cathieyun:
discussed in the issue here: https://github.com/dalek-cryptography/bulletproofs/issues/107


cathieyun:
Closing this PR because we discussed in https://github.com/dalek-cryptography/bulletproofs/issues/107 that padding circuits is not actually the best way to do this. We will instead just enforce that all circuit inputs have to follow that `circuit.n` is 0 or a power of 2.

Issue #118, opened 2018-07-18
R1CS support for circuits


cathieyun:
1. get r1cs → bp to work, using naive transformation https://github.com/dalek-cryptography/bulletproofs/pull/106
1.1 make sure r1cs → bp pads circuits if `n` is not 0 or a power of 2.
2. improve transformation to get rid of multiplications (simplify if only scalar mults)
3. improve r1cs → bp by bypassing matrices: https://github.com/dalek-cryptography/bulletproofs/pull/131

other tasks:
- benchmarking
- explore bellman <-> our r1cs
- convert bp to use `ProverInput` and `VerifierInput`
- move the arithmetic circuit code to a new file, so `mod` is more balanced

cathieyun:
One thing we would like to do is to give the verifier the ability to check that the circuit is correctly formed. This is easy for circuits that don't incorporate challenges (e.g. simple multiplication or addition: can check that the circuit is correctly formed easily.) For circuits such as merge/split and shuffle circuits that incorporate challenges, this is not as straightforward because the verifier should be able to check that the challenges were generated honestly (as opposed to being chosen maliciously). How should we do this?

1. The prover generates the challenges by committing to the variables, and using `ProofTranscript` to generate a challenge variable. This way, the verifier can receive the commitments to the variables and also use `ProofTranscript` to generate a challenge variable, and check that the challenge variable is the same.
2. Have a generic API/struct for a circuit builder parameterized by T, where T is either a pair of a secret and its commitment (if you are the prover) or just the commitment (if you are the verifier). Then, both the prover and verifier could build up a circuit from scratch, given the information they have. They can both generate challenge values (same as explained in point 1), and then can use those challenge values to `allocate_variable(T)`. Both the prover and verifier can `create_circuit()`, and should generate the same circuit if they have the same constraints and challenge variable.
3. Another solution?

I will start by implementing point 1. This is the most simple version, as it uses a provably-honestly-generated challenge value. The downside to this is that there isn't an elegant workflow for the verifier to check that the circuit is constructed honestly.

hdevalence:
I think that this problem would go away if we finished the R1CS-to-proof work first; then the prover and verifier can use the transcript to construct a circuit with the challenge values already built in.

cathieyun:
Thoughts on the `R1CS` representation:
[...]

PR #123, opened 2018-07-23
Pad circuits when creating them from `r1cs` format

cathieyun:
Makes sure r1cs → bp conversion pads circuits if `n` is not 0 or a power of 2.
A follow-up to PR https://github.com/dalek-cryptography/bulletproofs/pull/120 which enforces that inputs to circuit proofs must have `n` be 0 or a power of 2.

Discussed in this issue: https://github.com/dalek-cryptography/bulletproofs/issues/118

### DIFF
--- a/src/circuit_proof/r1cs.rs
+++ b/src/circuit_proof/r1cs.rs
@@ -18,7 +18,7 @@ pub struct Variable(usize);
 
 /// Represents a linear combination of some variables multiplied with their scalar coefficients,
 /// plus a scalar. E.g. LC = variable[0]*scalar[0] + variable[1]*scalar[1] + scalar
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub struct LinearCombination {
     variables: Vec<(Variable, Scalar)>,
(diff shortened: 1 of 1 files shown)


src/circuit_proof/r1cs.rs
```diff
@@ -167,11 +168,30 @@ impl ConstraintSystem {
...
+        mut self,
         pedersen_generators: &PedersenGenerators,
         rng: &mut R,
     ) -> (Circuit, ProverInput, VerifierInput) {
+        // If `n`, the number of multiplications, is not 0 or 2, then pad the circuit.
+        let n = self.a.len();
+        if !(n == 0 || n.is_power_of_two()) {
+            let pad_length = n.next_power_of_two() - n;
+            self.a.append(&mut vec![
+                LinearCombination::new(vec![], Scalar::zero());
```
hdevalence: Maybe there should be a `LinearCombination::zero()` constructor for this purpose?

cathieyun: done!

PR #125, opened 2018-07-24
Merge RP error-handling into the circuit branch.

hdevalence:
This doesn't change the circuit code to use error types (it leaves them as `&'static str` errors) because the circuit code is still in flux and I think that some of the possible errors could disappear if we changed the circuit code.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ byteorder = "1.2.1"
 serde = "1"
 serde_derive = "1"
 tiny-keccak = "1.4.1"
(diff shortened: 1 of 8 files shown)


cathieyun:
@hdevalence do you know why the PR passed all checks, but then broke the tests in the circuit branch (#109)? The tests say that the tests just ran for too long.

hdevalence:
I'm not sure, that looks like a spurious failure to me.

cathieyun:
ok!

PR #126, opened 2018-07-26
Make separate circuit module

cathieyun:
Split circuit code into its own module (instead of being in `circuit_proof/mod.rs`).

A nice side effect is, this forces us to be more deliberate with the circuit API, as the struct internals are no longer accessible from the `r1cs` module. So, I also added functions to the API for constructing structs.

Also, I changed the circuit creation and verification to use the `ProverInput` and `VerifierInput` structs, which makes the API cleaner.


cathieyun:
@hdevalence I tried to make this PR easier to review by copying `circuit_proof/mod.rs` to `circuit_proof/circuit.rs` so you could see the line deltas, but I don't think the github UI is picking up on that. Is there a better way to see the line deltas (the things I actually added/changed?)

hdevalence:
Hmm, it looks like there's just one commit, so not really. I think the way to achieve what you want would be to do one commit that does a `git mv` (but leaves the source tree in a broken state due to the moving), and then another commit on top of that one that makes all the changes you would need in order to reconcile the source tree back into a working state.

cathieyun:
Sounds good, I'll do that!

cathieyun:
Moving to a new PR and closing this one:
https://github.com/dalek-cryptography/bulletproofs/pull/130

hdevalence:
If you want, you can overwrite the branch with the new commits instead of making a new PR, so that all the discussion stays in one place.

Issue #110, opened 2018-07-11
Rework `Generators`


hdevalence:
The current `Generators` struct is designed to work for `m`-aggregated range proofs of `n` bits each, with both `m` and `n` powers of 2.  The aggregated rangeproof generators are constructed as a single chain of size `m*n`.

To deterministically construct nothing-up-my-sleeve generators, we can produce arbitrary-length chains of generators from a label by hashing the label to a point, then hashing that point to a new point, ....

For the circuit context, it may not be convenient to specify up-front how many generators will eventually be used, especially if we're building up R1CS systems in a bellman-like way. It seems a lot more convenient to just take the head of a lazily-evaluated arbitrary-length sequence of generators.

To make this more consistent with the rangeproofs, we could change the rangeproof generators to be `m` chains of `n` points each, with a different label for each `m`. The verifier would then join each of the `m` chains during verification.  This might also allow "rearrangable aggregation": currently, each party's vectors are concatenated, and the entries are combined with a giant vector of powers of a random factor, so each party needs to know their index in order to determine their place in that vector.  But maybe each party could just use a different `z` value or w/e, and the verifier could stitch these vectors together in the same way.  This would allow each party's proof shares to be freely arranged (allowing stuff like blind aggregation, aggregation with variable sizes, ... etc).

hdevalence:
It turns out that the rearrangeable aggregation idea doesn't work, because in order for each party to compute their shares independently, they need to commit their commitments to their own transcript (rather than have the dealer commit the sum of their commitments to the dealer's transcript). This means that the verifier needs to have each party's commitments (to recompute the each party's independent challenges), which makes the proofs linear instead of log-sized.

hdevalence:
The rest of this idea is handled already in #128, so I'm going to close this.

PR #128, opened 2018-07-30
Rework generators

hdevalence:
Partially addresses #110.

Now each party's generators are their own independent chain.


cathieyun:
Also `sha2` -> `sha3`

PR #129, opened 2018-07-30
Merge main into circuit to get reworked generators code

hdevalence:
This pulls:

- the reworked generators code (#128)
- the update to `curve25519-dalek` (#127)

PR #130, opened 2018-07-31
Make separate circuit module

cathieyun:
Split circuit code into its own module (instead of being in `circuit_proof/mod.rs`).

A nice side effect is, this forces us to be more deliberate with the circuit API, as the struct internals are no longer accessible from the `r1cs` module. So, I also added functions to the API for constructing structs.

Also, I changed the circuit creation and verification to use the `ProverInput` and `VerifierInput` structs, which makes the API cleaner.


cathieyun:
@hdevalence I followed your advice - if you look at the last commit ("clean up circuit API"), that contains all the diffs of interest.

hdevalence:
I think this is going to conflict with #129; let's merge this one first and then I'll redo the fixups from #129?

PR #131, opened 2018-07-31
R1CS -> BP without using matrices [WIP]

cathieyun:
TODO: update `LinearConstraint` so it uses a commitment to 1 instead of requiring a separate scalar?


src/circuit_proof/examples.rs
```diff
@@ -0,0 +1,247 @@
...
+
+    pub fn make_prover_cs(
+        &self,
+        type_0: Scalar,
+        type_1: Scalar,
+        val_in_0: Scalar,
+        val_in_1: Scalar,
+        val_out_0: Scalar,
+        val_out_1: Scalar,
+    ) -> (ConstraintSystem, Scalar) {
```
hdevalence: If the merge block is a circuit gadget, shouldn't it take a `&mut ConstraintSystem` as an input rather than returning one that it created? Since we don't have a way to combine constraint systems (and I don't think there's a good way to do that), doesn't constructing internally and returning the `ConstraintSystem` prevent joining two merge gadgets?

cathieyun: yeah that definitely makes more sense - fixed!

cathieyun:
Closing: moving work to this PR:
https://github.com/dalek-cryptography/bulletproofs/pull/138

PR #133, opened 2018-08-14
{WIP] Use "Merlin" transcripts

hdevalence:
This is still WIP, but submitted for comments/feedback.

Pre-merge checklist:

- [ ] release `0.1` of `merlin` to crates.io
- [ ] change `Cargo.toml` to not use a path dep
- [ ] add docs to TranscriptExt


src/transcript.rs
```diff
@@ -0,0 +1,36 @@
...
+        self.commit(b"sc", scalar.as_bytes());
+    }
+
+    fn commit_point(&mut self, point: &CompressedRistretto) {
+        self.commit(b"pt", point.as_bytes());
+    }
+
+    fn challenge_scalar(&mut self) -> Scalar {
+        let mut buf = [0u8; 64];
+        self.challenge(b"sc", &mut buf);
```
oleganza: `b"ch"`? :-)

cathieyun: sc = "scalar challenge"?

cathieyun:
This looks pretty good so far, I'm a fan of the `commit_point` and `commit_scalar` API!

src/transcript.rs
hdevalence: `sc` = scalar; we don't need to signal the commit/challenge distinction because it's already signaled by the operation type (which strobe commits to the transcript)

hdevalence:
Update: I published 0.1 of `merlin` today; I'll update this PR soon.

hdevalence:
I'm going to wait a bit to fix up this PR to also add the `ProverRng` functionality to Merlin; then this change can also get rid of the RNG parameter to all the prover functions.

cathieyun:
Just to check my understanding - in the latter setup, each prover has its own transcript, so the job of the MPC protocol (whether it's the dealer, or the parties passing messages to each other) is to make sure that each party has each of the other parties' commitments so that each party can add all the other parties' commitments to their transcript & generate challenges, right?

It seems that the latter setup would have a lot more communication overhead (if I understand it correctly), and the setup where the dealer holds the transcript seems much more efficient.

hdevalence:
Sorry, I think I was unclear: what I meant was that, instead of just using `thread_rng()` to generate randomness,  each `Party` would initialize a "personal" transcript internally, commit all the public data seen by that party to its personal transcript, and use the personal transcript to generate whatever randomness it needed for blinding factors, etc.  This way, the provers get to take advantage of the https://docs.rs/merlin/0.2.0/merlin/struct.TranscriptRng.html design (but without changing any of the communication).

cathieyun:
Thanks for the clarification! I don't have a strong opinion on whether the provers use `thread_rng()` or the `TranscriptRng` for generating randomness - it seems fine to keep using `thread_rng()` for now.

hdevalence:
Ok, let's keep using `thread_rng` and maybe swap it out later -- it's only on the prover's side, so we can freely change it later.

Issue #108, opened 2018-07-11
Check range proof assumptions about generator lengths


cathieyun:
Currently, the range proof module uses the `Generators`, and may be making assumptions that `Generators` enforces that `n` and `m` are powers of two. However, there are two problems with this - the `Generators` file doesn't actually perform any checks on `n` and `m`, and for circuit proofs we want to use the `Generators` with `n` (and maybe also `m`) that is not a power of 2. 

TODO: go through the range proof code and find where it is assuming that `Generators` enforces that `n` and `m` are powers of two, and do those checks in the range proof itself. 

Initially discussed here: https://github.com/dalek-cryptography/bulletproofs/pull/102

PR work here:  https://github.com/dalek-cryptography/bulletproofs/pull/134

cathieyun:
Merge #134!

PR #134, opened 2018-08-15
Check generator lengths in range proof

cathieyun:
Issue here: https://github.com/dalek-cryptography/bulletproofs/issues/108

Basically, when doing range proofs, check that `Generators` has `n` and `m` as powers of 2.


cathieyun:
PTAL?

PR #138, opened 2018-08-29
R1CS rework

cathieyun:
Change R1CS API to allocate all variables (high-level _and_ low-level variables), use them in `LinearCombination`s, and then add those `LinearCombination`s to a `ConstraintSystem`. 

The next step will be to have a `prove()` and `verify()` function directly on an `R1CS` instance. (In this PR, we use the `circuit` module for that.) To keep PRs small and manageable, that will be in a separate PR.


cathieyun:
PTAL

hdevalence:
Is it OK to merge some changes from `main` into `circuit` and then into this branch before reviewing them?

src/circuit_proof/r1cs.rs
```diff
@@ -1,26 +1,38 @@
 #![allow(non_snake_case)]
```
oleganza: Light bikeshed: how about a friendlier name than `r1cs`? I just remember how it's intimidating when very specialized terms are used in things I'm trying to learn, making it harder for now good reason. E.g. it's not like we have rank-0 and rank-2 constraint systems too, so we need to distinguish between them.

How about `constraints.rs`?

oleganza: @hdevalence yeah, good point. My original comment relates to reviewing the code as a whole, so I think it still applies regardless of whether this module is exposed to the users.

src/circuit_proof/r1cs.rs
```diff
@@ -53,112 +65,154 @@ impl LinearCombination {
...
+            VariableType::v => {
+                self.v_assignment.push(value);
+                self.v_assignment.len() - 1
+            }
+        };
+        Variable { var_type, index }
+    }
+
+    // Get number of multiplications.
+    pub fn get_n(&self) -> usize {
```
oleganza: Rename to `multipliers_count` and remove the comment? :-)

src/circuit_proof/r1cs.rs
```diff
@@ -53,112 +65,154 @@ impl LinearCombination {
...
+        let n = self.aL_assignment.len();
+        if n == 0 || n.is_power_of_two() {
+            return n;
+        }
+        return n.next_power_of_two();
+    }
+
+    // Get number of high-level witness variables.
+    pub fn get_m(&self) -> usize {
```
oleganza: Maybe `commitments_count` (if we try to consistently associated Vs with "commitments" and treat A,S as implementation detail and part of the proof)?

src/circuit_proof/r1cs.rs
```diff
@@ -53,112 +65,154 @@ impl LinearCombination {
...
+        self.v_assignment.len()
+    }
+
+    pub fn make_V(
```
oleganza: This feels a little bit out-of-order. In principle, V commitments exist outside the circuit proof. Proof just need the components of these commitments to compute things, but it's not the _source_ of the commitments. So taking in blinding factors and then reading values from the wires to form commitments seems like reversed relationship.

What if the proving system takes something like `OpenCommitment` which can give you a Scalar value and Scalar blinding factor, and it constructs the proof. But "closing" the commitment will happen outside the proof - e.g. in the scope of the outer protocol. E.g. when you create a transaction, you send your secret values to the proving system, receive the proof, then commit your secrets and store them in the proper slots in the transaction, but outside the proof. The proof is stored (maybe aggregated) elsewhere in the transaction.

src/circuit_proof/r1cs.rs
```diff
@@ -53,112 +65,154 @@ impl LinearCombination {
...
             .iter()
             .zip(v_blinding)
-            .map(|(v_i, v_blinding_i)| pedersen_gens.commit(*v_i, *v_blinding_i))
-            .collect();
-        VerifierInput::new(V)
+            .map(|(v_i, v_blinding_i)| Ok(pedersen_gens.commit(v_i.clone()?, *v_blinding_i)))
+            .collect()
+    }
+
+    pub fn constrain(&mut self, lc: LinearCombination) {
```
oleganza: Very subjective nitpick: `add_constraint` would read more clearly to me at the call site.

src/circuit_proof/r1cs.rs
```diff
@@ -53,112 +65,154 @@ impl LinearCombination {
     }
 }
 
-/// Represents a vector of groups of 3 linear combinations, where a * b = c
 pub struct ConstraintSystem {
-    // a[i] * b[i] = c[i] for all i
-    a: Vec<LinearCombination>,
-    b: Vec<LinearCombination>,
-    c: Vec<LinearCombination>,
+    lc_vec: Vec<LinearCombination>,
```
oleganza: `lc_vec` -> `constraints`?

hdevalence:
Merged in the latest `main`/`circuit` changes into this branch.

src/errors.rs
```diff
@@ -88,3 +88,23 @@ pub enum MPCError {
...
+
+/// Represents an error during the proving or verifying of a constraint system.
+#[derive(Fail, Clone, Debug, Eq, PartialEq)]
+pub enum R1CSError {
+    // TODO: make better errors
+    #[fail(display = "Invalid R1CS construction.")]
+    InvalidR1CSConstruction,
+    // When trying to access a variable that has an Err as its value assignment
+    #[fail(display = "Variable does not have a value assignment.")]
+    InvalidVariableAssignment,
```
hdevalence: Is the InvalidVariableAssignment ever used aside from the case where the verifier doesn't have the witness data to fill in?

If it isn't, should we rename it to be MissingAssignment?

If it is, should we split the error types?

src/circuit_proof/r1cs.rs
cathieyun: @oleganza that makes sense. I just think that `constraints` is too generic... whereas `constraint_system` is a better description of what the code is creating. /shrug both are better than r1cs though

hdevalence: Just my 2c, but I think `r1cs` is fine, it's short, unique, and a specific term of art used by other systems (not just something we made up).

src/errors.rs
cathieyun: It's only used when the verifier doesn't have witness data.
Sure, we can rename it to MissingAssignment!

src/circuit_proof/r1cs.rs
```diff
@@ -53,112 +58,138 @@ impl LinearCombination {
...
-    // a[i] * b[i] = c[i] for all i
-    a: Vec<LinearCombination>,
-    b: Vec<LinearCombination>,
-    c: Vec<LinearCombination>,
+    constraints: Vec<LinearCombination>,
 
-    // Assignments of variables
-    var_assignment: Vec<Scalar>,
+    // variable assignments
+    aL_assignment: Vec<Assignment>,
```
oleganza: Maybe `*_assignments` in plural?

src/circuit_proof/circuit.rs
```diff
@@ -128,6 +128,35 @@ pub struct CircuitProof {
...
+            T_3,
+            T_4,
+            T_5,
+            T_6,
+            t_x,
+            t_x_blinding,
+            e_blinding,
+            ipp_proof,
+        }
+    }
```
hdevalence: Where does this function get used? I am guessing there's a reason that the `CircuitProof` can't just be constructed directly?

src/circuit_proof/r1cs.rs
```diff
@@ -1,26 +1,31 @@
...
-// This is a stripped-down version of the Bellman r1cs representation, for the purposes of
-// learning / understanding. The eventual goal is to write this as a BulletproofsConstraintSystem
-// that implements the Bellman ConstraintSystem trait, so we can use that code/logic.
-// (That would require the bellman code to be decoupled from the underlying pairings.)
+pub type Assignment = Result<Scalar, R1CSError>;
 
-/// Represents a variable in our constraint system, where the value represents the index.
-#[derive(Copy, Clone, PartialEq, Eq, Debug)]
-pub struct Variable(usize);
+pub fn err_assignment() -> Assignment {
```
hdevalence: Maybe this could just be
```
pub fn missing() -> Assignment { ... }
```

src/circuit_proof/r1cs.rs
```diff
@@ -1,26 +1,31 @@
...
+    Err(R1CSError::InvalidVariableAssignment)
+}
 
 /// Represents a linear combination of some variables multiplied with their scalar coefficients,
-/// plus a scalar. E.g. LC = variable[0]*scalar[0] + variable[1]*scalar[1] + scalar
+/// plus a scalar. The linear combination is supposed to evaluate to zero.
+/// E.g. LC = 0 = variable[0]*scalar[0] + variable[1]*scalar[1] + scalar
```
hdevalence: Is evaluating to zero really a property of the `LinearCombination`, or of the way that the `ConstraintSystem` will use the linear combinations?

hdevalence:
As a kind of general question, in this PR there's still two modules, one for circuits and one for the constraint system. The constraint system builds up some circuit data, which is used by the prover and verifier.

But it should be possible to go directly from the constraint system data to the proofs, and it seems like this would make the code a lot simpler -- the circuit code could be dropped entirely -- and it would also fit better with stuff like dynamic circuits using transcript challenges.

src/circuit_proof/r1cs.rs
```diff
@@ -53,112 +58,138 @@ impl LinearCombination {
...
     }
 
-    fn create_verifier_input(
+    pub fn multipliers_count(&self) -> usize {
+        let n = self.aL_assignment.len();
+        if n == 0 || n.is_power_of_two() {
+            return n;
+        }
+        return n.next_power_of_two();
+    }
```
hdevalence: Maybe this should be
```
fn innerproduct_size(&self) -> usize {
    self.aL_assignment.len().next_power_of_two()
}
```
since what this function returns isn't actually the number of multipliers, but the size of the inner product vectors?

Except for the `n == 0` case, it's not necessary to check whether `n` is already a power of two, and I think we can probably exclude `n == 0` in this new setting because `n` will be zero only when we have an empty constraint system, which shouldn't (imo) be allowed.

src/circuit_proof/circuit.rs
cathieyun: I forgot to remove this after it was no longer needed, will clean up

src/circuit_proof/r1cs.rs
cathieyun: The latter - I will clarify that in the comment.

src/circuit_proof/r1cs.rs
cathieyun: Agreed - I'll deleted the `make_V` function, and only have the functionality for "making V commitments" exposed in the `make_proof_input` function, which is what you call to create the `circuit`, `prover_input`, and `verifier_input` (so, it seems helpful and fitting there). In future changes (like in the next PR, where we get rid of the `make_proof_input` workflow and just have a `prove()` and `verify()` function on r1cs, we can get rid of the commitment-forming logic altogether.

src/circuit_proof/r1cs.rs
cathieyun: Alright let's keep `r1cs` for now :)

cathieyun:
I'd rather do the R1CS `prove()/verify()` work in another PR (just to keep PR sizes small and manageable, since this one is already a big change). So, let's just merge this as-is?

oleganza:
@cathieyun 👏

PR #139, opened 2018-08-31
Merge main into circuit

hdevalence:
This should be mostly rubber-stamp-able, the only substantive change is rewriting the circuit methods to use Merlin transcripts.


src/transcript.rs
```diff
@@ -0,0 +1,62 @@
...
+fn le_u64(value: u64) -> [u8; 8] {
+    let mut value_bytes = [0u8; 8];
+    LittleEndian::write_u64(&mut value_bytes, value);
+    value_bytes
+}
+
+impl TranscriptProtocol for Transcript {
+    fn rangeproof_domain_sep(&mut self, n: u64, m: u64) {
+        self.commit_bytes(b"dom-sep", b"rangeproof");
+        self.commit_bytes(b"n", &le_u64(n));
```
oleganza: Would it be cleaner to avoid `"dom-sep"` and instead do something like this?

```
fn rangeproof_domain(n,m) {
    commit("RangeProof.n", n)
    commit("RangeProof.m", m)
}

fn innerproduct_domain(n) {
    commit("InnerProductProof", n)
}
```

hdevalence: I think it's cleaner to keep them as different messages, since it means there's one message to ID the protocol, plus as many other messages as needed to fill in the parameters.

PR #140, opened 2018-08-31
Proving and verifying directly on R1CS instance

cathieyun:
This PR contains:
- API modifications to prove and verify directly on the R1CS instance. This means you can do `proof = ConstraintSystem::prove(...)` and `ConstraintSystem::verify(proof, ...)`
- Changes to the `Assignment` struct so that you can do arithmetic and try operations on it. This makes it easier to implement the same workflow for the Prover and Verifier when they are building their `ConstraintSystem`.
- Making some more modules public so that they can be used in cryptonumia.

Note: currently, the `R1CS` module still uses the `circuit` module internally, to create the `circuit` from the `R1CS` representation (which is used in the proving and verifying steps). In future changes, we can get rid of this dependency on the `circuit` module (by moving the logic to the `R1CS` module instead). However, I am not doing that in this PR because my goal in this PR is to create a more finalized API (so I can use it for cryptonumia)... the other changes can be done later without affecting the API at all.


src/circuit_proof/r1cs.rs
```diff
@@ -182,79 +215,73 @@ impl ConstraintSystem {
...
         let circuit = self.create_circuit();
-        let prover_input = self.create_prover_input(&v_blinding);
-        let verifier_input = self.create_verifier_input(pedersen_gens, &v_blinding);
+        let prover_input = self.create_prover_input(&v_blinding)?;
+        let verifier_input = self.create_verifier_input(&gen.pedersen_gens, &v_blinding)?;
 
-        (circuit, prover_input, verifier_input)
+        // TODO: use error handling instead of unwrap
+        let circuit_proof =
+            CircuitProof::prove(&gen, transcript, rng, &circuit, &prover_input).unwrap();
```
oleganza: maybe add a `map_err` right away since outer functions already returns Result?

oleganza: Maybe add a `map_err` right away since the outer function already returns Result and avoid the `unwrap`?

src/circuit_proof/assignment.rs
```diff
@@ -0,0 +1,135 @@
...
+// The assignment value to a variable, as stored in `ConstraintSystem`.
+// Provers create a `Value` assignment, while verifiers create an `Missing` assignment.
+#[derive(Copy, Clone, Debug)]
+pub enum Assignment {
+    Value(Scalar),
+    Missing(),
+}
+
+impl Assignment {
+    pub fn new(scalar: Scalar) -> Self {
```
hdevalence: Could this be `impl From<Scalar> for Assignment`?

cathieyun: yeah!
I also implemented From<u64> while I was at it, thanks for the suggestion

src/circuit_proof/assignment.rs
```diff
@@ -0,0 +1,135 @@
...
+        Assignment::Value(scalar)
+    }
+
+    pub fn zero() -> Self {
+        Assignment::Value(Scalar::zero())
+    }
+
+    pub fn from_u64(v: u64) -> Self {
```
hdevalence: This seems like it could also be a `From` impl -- I guess it probably makes sense to provide multiple From impls instead of making users do two `from`s in a row?

PR #143, opened 2018-09-10
Update R1CS API

cathieyun:
- require user to pass in a `Transcript` and `Commitments` to `ConstraintSystem::new()`, such that:
   - `ConstraintSystem` holds onto its own `Transcript`,
   - `ConstraintSystem` takes the `Commitments` and inputs them into the `Transcript`, which should be the only time that the `Transcript` takes inputs from that `ConstraintSystem` instance
- add a `get_challenge` function to `ConstraintSystem` so that gadgets can get challenge variables from the `Transcript`
- make `Variable` `Copy`-able

### DIFF
--- a/src/circuit_proof/r1cs.rs
+++ b/src/circuit_proof/r1cs.rs
@@ -57,8 +57,8 @@ impl LinearCombination {
     }
 }
 
-pub struct ConstraintSystem {
-    transcript: Transcript,
+pub struct ConstraintSystem<'a> {
+    transcript: &'a mut Transcript,
     constraints: Vec<LinearCombination>,
 
     // variable assignments
@@ -68,9 +68,9 @@ pub struct ConstraintSystem {
     v_assignments: Vec<Assignment>,
 }
 
-impl ConstraintSystem {
+impl<'a> ConstraintSystem<'a> {
     pub fn prover_new(
-        transcript: Transcript,
+        transcript: &'a mut Transcript,
         // TODO: encapsulate v, v_blinding, pedersen_gens into one "commitment" struct
         v: Vec<Scalar>,
         v_blinding: Vec<Scalar>,
@@ -106,7 +106,7 @@ impl ConstraintSystem {
     }
 
     pub fn verifier_new(
-        transcript: Transcript,
+        transcript: &'a mut Transcript,
         commitments: Vec<RistrettoPoint>,
     ) -> (Self, Vec<Variable>) {
         let mut cs = ConstraintSystem {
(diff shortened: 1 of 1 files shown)


src/circuit_proof/r1cs.rs
```diff
@@ -57,6 +58,7 @@ impl LinearCombination {
 }
 
 pub struct ConstraintSystem {
+    transcript: Transcript,
```
hdevalence: This is taking ownership of the `Transcript` object directly, but I think we want to take ownership just of a mutable borrow of a `Transcript` owned by some higher-level function.  So it seems like the definition should be

```rust
pub struct ConstraintSystem<'a> {
    transcript: &'a mut Transcript,
    /* ... */
}
```

src/circuit_proof/r1cs.rs
```diff
@@ -67,8 +69,13 @@ pub struct ConstraintSystem {
 }
 
 impl ConstraintSystem {
-    pub fn new() -> Self {
+    pub fn new(mut transcript: Transcript, commitments: Vec<RistrettoPoint>) -> Self {
```
hdevalence: then this would become
```rust
impl<'a> ConstraintSystem<'a> {
    pub fn new(transcript: &'a mut Transcript, /*... */) -> { /* ... */ }
}
```

src/circuit_proof/r1cs.rs
```diff
@@ -138,6 +145,10 @@ impl ConstraintSystem {
         self.constraints.push(lc);
     }
 
+    pub fn get_challenge(&mut self, label: &'static [u8]) -> Scalar {
```
hdevalence: This could be just `challenge_scalar` to match the `TranscriptProtocol` API

cathieyun:
Updated:
- `ConstraintSystem` takes in a mutable borrow of a `Transcript`
- `ConstraintSystem` has `challenge_scalar()` function

src/circuit_proof/r1cs.rs
```diff
@@ -66,15 +68,64 @@ pub struct ConstraintSystem {
...
+        v_blinding: Vec<Scalar>,
+        pedersen_gens: PedersenGenerators,
+    ) -> (Self, Vec<Variable>, Vec<RistrettoPoint>) {
+        let mut cs = ConstraintSystem {
+            transcript,
+            constraints: vec![],
+            aL_assignments: vec![],
+            aR_assignments: vec![],
+            aO_assignments: vec![],
+            v_assignments: vec![],
```
hdevalence: Where does `v_assignments` get initialised?

hdevalence: Oh, does that happen via the calls to `assign_committed`?

cathieyun: Here, the `v_assignments` gets initialized in line 102, in the `cs.assign_committed()` call

hdevalence:
This looks great!

Issue #142, opened 2018-09-10
Remove the `circuit` module and work solely with the constraint system


hdevalence:
PR #140 made proving and verifying work on the constraint system.  The next step is to entirely remove the circuit module, move whatever code is still needed into the constraint system implementation, and pass directly from the constraints to proof creation or verification.

hdevalence:
In the rangeproof code, we orient the whole API around the proof struct, which has prove and verify methods.

However, for the circuit proofs, we have two choices for how to orient the prove/verify API: around the constraint system, or around the proof struct.

In the first case, we would have `prove` and `verify` methods on the `ConstraintSystem`, which would either produce a proof or a verification result.

In the second case, we would retain the `CircuitProof` (or maybe call it `R1CSProof`), and it would have `prove` and `verify` methods which would take a `ConstraintSystem` as a parameter.  (In both cases, we can consume the `ConstraintSystem` to prevent reuse).

Currently, I'm leaning towards the second option, because it lets us provide an API with two distinct phases (building the constraint system, and then using it to generate or verify a proof), and it matches the API we have for rangeproofs already -- as in the rangeproof API, proof creation and verification are done on the proof structs, and the additional complexity of having to specify the statement to be proved is done as an additional step.

Also, if the `ConstraintSystem` takes ownership of the transcript reference, then we don't need to pass a transcript to the prove/verify calls, since it's already included in the constraint system.

WDYT @cathieyun?

cathieyun:
The current implementation uses the first case, but both seem fine to me. If I understand correctly, the distinction between the two is just whether you call `verifier_cs.verify(proof, ...)` or `proof.verify(verifier_cs, ...)` right? (where both cases generate a proof with `proof = prover_cs.prove(...)`, and both cases consume `prover_cs` in the verification step).

This doesn't seem like a huge difference, so if the second case is more in line with the `RangeProof` API, then I'm on board with using that.

hdevalence:
Yeah, it's exactly that distinction, and it's not a huge difference either way. I'll work on the second case!

hdevalence:
I realized that that approach won't work, because proving and verification requires access to the constraint system internals. So it'll be a method on the constraint system.

hdevalence:
Starting to work on this now that #143 is merged.

hdevalence:
Closed by #145

PR #145, opened 2018-09-11
Eliminate the `circuit` module and prove directly from the constraint data

hdevalence:
WIP branch for #142


src/transcript.rs
```diff
@@ -12,6 +12,8 @@ pub trait TranscriptProtocol {
     fn innerproduct_domain_sep(&mut self, n: u64);
     /// Commit a domain separator for a circuit.
     fn circuit_domain_sep(&mut self, n: u64, m: u64, q: u64);
```
hdevalence: Oh, I guess this function should go away

Issue #150, opened 2018-09-13
Generators should be changed to have a notion of capacity


hdevalence:
For circuit proofs, we stream constraints into the constraint system, so it's not known in advance how many generators we're going to need. The current `Generators` struct has a fixed number of generators for each party, and the code checks that this is equal to the number of generators needed.

However, since we already changed the `Generators` struct to generate each aggregation party's generators using a different, domain separated XOF, we can view each party's share in the `Generators` struct as the leading slice of an arbitrarily-long chain of independent generators.

This gives a nice way to solve the problem for the constraint system: we can change the `Generators` struct to have a notion of a `capacity` (the length of the slice), construct `capacity` many generators for each party on initialization, and have all of the proving methods check that the number of generators they need is less than the capacity (or else return an error). 

Checklist:

- [x] Add a notion of `capacity` to the `Generators` struct
- [x] Change the rangeproof code to be capacity-aware
- [x] Merge those changes into `main`
- [x] Merge `main` into `circuit` and update the circuit code to be capacity-aware.

hdevalence:
The first three points are done by #152

cathieyun:
Last point is done by https://github.com/dalek-cryptography/bulletproofs/pull/153, so I'm closing this issue

PR #151, opened 2018-09-13
Split ConstraintSystem into ProverCS, VerifierCS, and a trait

### LINKED ISSUE #149: Split constraint system logic into prover/verifier types with a common trait
Instead of having a single `ConstraintSystem` type, we should experiment with having

```rust

trait ConstraintSystem;

struct ProverCS;
struct VerifierCS;

impl ConstraintSystem for ProverCS {};
impl ConstraintSystem for VerifierCS {};
```

so that it's clear which code is supposed to abstract over both roles (e.g., gadgets) and which code is supposed to only prove or verify (so it's impossible to call verify on a prover, for instance).

hdevalence:
Closes #149 

I didn't keep all of the tests, just the most complex one, because there's more changes to make after this, and it seemed better to wait until after the API is more settled to add more of the tests back.

This API isn't really usable yet since it removes the ability to externally count the number of generators needed, but doesn't have a way to have "enough" generators (which would be fixed by #150)


cathieyun:
Want to put the `LinearCombination` and `ConstraintSystem` definitions into the `mod.rs` file, so we don't need the `r1cs` file anymore?

hdevalence:
Sure, that sounds like a good idea.

PR #152, opened 2018-09-13
Add a notion of capacity to the Generators struct

hdevalence:
For rangeproofs, this allows building a single Generators object and using it for any aggregation or bitsize.

For circuit proofs, this means the user doesn't have to predetermine the number of constraints in their circuit.


hdevalence:
Progress towards #150

src/generators.rs
```diff
@@ -164,24 +248,46 @@ mod tests {
     extern crate hex;
     use super::*;
 
+    // XXX write tests
```
cathieyun: can we remove this now?

PR #154, opened 2018-09-14
 Replace constant terms in LC by constant Variables 

hdevalence:
This replaces the constant term in the `LinearCombination` struct by adding a new `Constant` variant to the `Variable` enum which represents 1.

I changed the example gadget to use it by changing `c2` to be a constant.  I'm not sure whether this is the best possible API yet. WDYT?

### DIFF
--- a/src/circuit_proof/mod.rs
+++ b/src/circuit_proof/mod.rs
@@ -55,7 +55,7 @@ pub enum Variable {
     /// Represents the output of a multiplication gate.
     MultiplierOutput(usize),
     /// Represents the constant 1.
-    Constant(),
+    One(),
 }
 
 /// Represents a linear combination of `Variables`.  Each term is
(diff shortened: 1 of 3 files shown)


cathieyun:
I think this looks good - for the end user, it's not that different from adding a constant to the end of the linear constraint, but it is a little cleaner. If the variable type is `Variable::One()` it would be less confusing; other than that, this PR LGTM

hdevalence:
Changed to `Variable::One`.

We might want to think a little bit about the privacy of the `Variable` internals, right now the CS users can access / modify the indices they hold.  I'm not sure if it's a big deal, since the R1CS api is already low-level and has lots of ways to shoot yourself in the foot just by writing the constraints wrong, so if someone wants to go out of their way to tamper with the variable indexes, maybe that's just their problem..

hdevalence:
Yeah, something like that. I think it's not possible to hide the variants of an enum without hiding the enum itself.  Right now someone who has a `Variable` could edit its internal index and pass it back into the CS, if they wanted.

PR #158, opened 2018-09-14
ConstraintSystems hold onto generators struct

### LINKED ISSUE #155: Constraint systems should hold onto a Generators object
I think it might be cleaner to pass in borrows of everything the constraint systems will need to prove and verify up front, so that the final `prove` / `verify` calls just consume the CS and don't need any further parameters.

cathieyun:
closes #155


cathieyun:
I updated the `verifier::prove()` function to make its own Rng from the transcript.
PTAL?

src/circuit_proof/verifier.rs
```diff
@@ -156,11 +153,19 @@ impl<'a> VerifierCS<'a> {
...
-        let gens = generators.share(0);
+        let gens = self.generators.share(0);
+
+        // Create a `TranscriptRng` from the transcript
+        let mut rng = {
+            let ctor = self.transcript.fork_transcript();
+
+            use rand::thread_rng;
+            ctor.reseed_from_rng(&mut thread_rng())
+        };
```
hdevalence: nit: I think this could be slightly cleaner as
```
use rand::thread_rng;
let mut rng = self.transcript.fork_transcript().reseed_from_rng(&mut thread_rng());
```
The prover's code is more complicated, so it bundles stuff up into a constructor scope, but I don't think that that's needed here.

cathieyun: 👍

Issue #157, opened 2018-09-14
Verification should operate on compressed points


hdevalence:
The verifier CS should operate on compressed points, and decompress its points right as it's feeding them into the verification equation.  This is probably easiest to do after #148.

cathieyun:
closed by https://github.com/dalek-cryptography/bulletproofs/pull/160

PR #160, opened 2018-09-14
Verification should operate on compressed points

cathieyun:
Addresses https://github.com/dalek-cryptography/bulletproofs/issues/157

I'm not sure how to have the verifier "decompress its points right as it's feeding them into the verification equation." - currently just decompressing as a separate step before the mega_check.

@hdevalence can we pair on inlining the decompression?


cathieyun:
inlined the point decompression in the R1CS verification function! PTAL

src/circuit_proof/verifier.rs
```diff
@@ -303,21 +282,21 @@ impl<'a> VerifierCS<'a> {
...
+            iter::once(Some(A_I))
+                .chain(iter::once(Some(A_O)))
+                .chain(iter::once(Some(W_L_point)))
+                .chain(iter::once(Some(W_R_point)))
+                .chain(iter::once(Some(W_O_point)))
+                .chain(iter::once(Some(S)))
+                .chain(iter::once(Some(gens.pedersen_gens.B)))
+                .chain(iter::once(Some(gens.pedersen_gens.B_blinding)))
+                .chain(gens.G(n).map(|&g_i| Some(g_i)))
+                .chain(gens.H(n).map(|&h_i| Some(h_i)))
```
hdevalence: These might be `G_i`, `H_i` if we're using capitals for the points and `g_i`, `h_i` for their scalars?

cathieyun: 👍

cathieyun:
Actually, it looks like we can inline some more decompression (of points that aren't used in any calculations before the mega_check). I'm working on that now!

PR #163, opened 2018-09-18
Disentangle Pedersen generators from Bulletproofs generators

### LINKED ISSUE #161: Split Pedersen commitments from Bulletproofs generators
The `Generators` struct has both generators for bulletproofs, as well as generators for the Pedersen commitments.  These should be seperate, since we have the flexibility to accept arbitrary PC generators, but we don't expose it in the API. 

Why does it matter that the PC generators are flexible? Because it's way more likely that the PC generators will change per-protocol, and there are applications (like algebraic mac credentials) that it would be good to be able to interoperate with.

hdevalence:
Closes #161.


hdevalence:
I also had to amend the commit to change the benchmarks to match the new API, so it was probably broken in at least two ways

src/range_proof/mod.rs
```diff
@@ -369,13 +371,12 @@ fn delta(n: usize, m: usize, y: &Scalar, z: &Scalar) -> Scalar {
...
     use super::*;
-    use rand::rngs::OsRng;
 
-    use generators::PedersenGenerators;
+    use generators::PedersenGens;
 
     #[test]
     fn test_delta() {
-        let mut rng = OsRng::new().unwrap();
+        let mut rng = rand::thread_rng();
```
cathieyun: what's the motivation for changing the source of randomness?

hdevalence: There's no really good reason -- I removed the `OsRng` that was used in other parts of the test code while changing that code, then removed the `use` declaration that brought it into scope, then realized that the `test_delta` function was using it, and didn't want to re-add it, so I just made everything use `thread_rng`

PR #165, opened 2018-09-24
Use Merlin v0.4

hdevalence:
Merlin 0.3->0.4 changes the `TranscriptRng` API.  In Bulletproofs, we only used that for the verifier's randomness, but that randomness can just use a `thread_rng` since there's no secrets to leak.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ serde = "1"
 serde_derive = "1"
 tiny-keccak = "1.4.1"
 failure = "0.1"
-merlin = "0.3"
+merlin = "0.4"
 
 [dev-dependencies]
 hex = "^0.3"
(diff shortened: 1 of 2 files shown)


oleganza:
Isn't it still safer to make that batching challenge not depend purely on RNG? The risk here is that verifier might be tricked to accepting (or rejecting) a proof when the opposite must happen. E.g. forcing verifier to fork off a blockchain state. Random challenge is safer than a fully deterministic one as the prover cannot predict it at all, but deterministic is still exponentially better than predictable due to failed RNG.

hdevalence:
It seemed simpler, with less components to rely on, I guess. But if the `TranscriptRng` is broken then the whole proof system is, so there's probably not a downside to using the `TranscriptRng` for verification.

hdevalence:
Changed back to using the `TranscriptRng`

PR #166, opened 2018-09-24
Implement PartialEq for Assignment

cathieyun:
This is extremely useful in the `cloak` gadgets, where the prover is comparing assignment values to determine the values of intermediate variables.

### DIFF
--- a/src/circuit_proof/assignment.rs
+++ b/src/circuit_proof/assignment.rs
@@ -1,8 +1,7 @@
 use curve25519_dalek::scalar::Scalar;
 use errors::R1CSError;
-use std::cmp::PartialEq;
 use std::ops::{Add, Div, Mul, Sub, Try};
-use subtle::ConstantTimeEq;
+use subtle::{Choice, ConditionallyAssignable, ConditionallySelectable, ConstantTimeEq};
 
 // The assignment value to a variable, as stored in `ConstraintSystem`.
 // Provers create a `Value` assignment, while verifiers create an `Missing` assignment.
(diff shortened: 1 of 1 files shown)


hdevalence:
Hmm. Do we want the prover to operate in constant time?  If we do, maybe we should implement the `subtle` traits rather than `PartialEq`.

I don't have a good idea what the right design is here. I think that Bellman doesn't attempt to be timing-safe at all.

cathieyun:
Due to `dalek` using subtle version `0.7.1`, but `bool from Choice` requiring subtle version `0.9`, it looks like we can't use the timing-safe equality check until `dalek` updates its version of `subtle`.

So, let's use non-constant-time equality checking for now, and update it to constant-time checking when `dalek` updates its version of `subtle`.

src/circuit_proof/assignment.rs
```diff
@@ -137,3 +139,15 @@ impl Try for Assignment {
...
+    fn eq(&self, other: &Assignment) -> bool {
+        match (self, other) {
+            (Assignment::Value(self_value), Assignment::Value(other_value)) => {
+                bool::from(self_value.ct_eq(other_value))
+            }
+            (Assignment::Missing(), Assignment::Missing()) => true,
+            _ => false,
+        }
+    }
+}
```
hdevalence: This is implementing `PartialEq` in constant time, but because `PartialEq` returns a `bool`, it's not really effective to do the check in constant time, because the `bool` that comes out is probably going to be used in an `if` statement, which wouldn't be constant time.

Instead I think it would be better to implement the `subtle`  `ConstantTimeEq` and `ConditionallySelectable` traits for `Assignment`, and then, maybe, implement `PartialEq` in terms of those. I'm not sure if it makes sense to have a `PartialEq` implementation for `Assignment`, or whether we should try to push people to write constant-time gadgets.

src/circuit_proof/assignment.rs
```diff
@@ -137,3 +138,32 @@ impl Try for Assignment {
...
+impl ConditionallySelectable for Assignment {
+    // This function should execute in constant time for a and b of the same type.
+    // So, if a and b are both of type Assignment::Value(), then the comparison will
+    // execute in constant time regardless of their value assignments.
+    fn conditional_select(a: &Self, b: &Self, choice: Choice) -> Self {
+        match (a, b) {
+            (Assignment::Value(a_val), Assignment::Value(b_val)) => {
+                let mut out_val = a_val.clone();
+                out_val.conditional_assign(&b_val, choice);
+                Assignment::from(out_val)
```
hdevalence: This could have a FIXME note with a reference to the upstream issue

hdevalence:
Good find about the `curve25519_dalek` trait implementations, thanks! We could merge this with a note to change the `ConditionallySelectable` implementation when it's fixed upstream.

src/circuit_proof/assignment.rs
cathieyun: done!

cathieyun:
Filed issues on `curve25519-dalek` and `subtle`:
https://github.com/dalek-cryptography/curve25519-dalek/issues/200
https://github.com/dalek-cryptography/subtle/issues/36

PR #169, opened 2018-09-25
Add documentation and doctests to rangeproof code.

### LINKED ISSUE #162: Add documentation to rangeproofs
The rangeproofs should have doc-tests that demonstrate how to use the API.

hdevalence:
This adds `#![deny(missing_docs)]`, so it will probably require changes to the `circuit` branch the next time we sync the two branches (since it will force documenting all of the circuit API).

Closes #162

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,11 +16,10 @@ curve25519-dalek = { version = "0.19", features = ["serde"] }
 subtle = "0.7"
 sha3 = "0.7"
 digest = "0.7"
-rand = "0.5.0-pre.2"
-byteorder = "1.2.1"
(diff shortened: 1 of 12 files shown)


hdevalence:
This also renames `ValueCommitment` to `BitCommitment` (so that we're not overloading between the `(V, A, S)` tuple and just the `V` point), which I think means that we should update the embedded diagram -- @oleganza do you have a copy of the original Sketch file?

oleganza:
@hdevalence the diagram source file is committed into `assets/` folder. I can change it on this branch if you'd like.

hdevalence:
That'd be great!

PR #175, opened 2018-09-26
Update diagram with BitChallenge/BitCommitment terms

oleganza:
Sketch file and generated SVG are updated to reflect new terminology introduced in #169.


oleganza:
Fixed. PTAL!

oleganza:
@hdevalence i've renamed remaining `value_{commitment,challenge}` into `bit_*`, PTAL again?

PR #179, opened 2018-10-01
Add API docs for circuit code

hdevalence:
Merges `main` into `circuit`, which pulls in `#![deny(missing_docs)]`, and fills in API documentation for the R1CS code.

This still lacks a doctest example for the `r1cs` module, which we should include to show how all the pieces fit together.  This should include a gadget function written generically over a `ConstraintSystem`, its use by a `ProverCS` for proving, and then verification using a `VerifierCS`.  It would be especially good to use all the API we have, including the `challenge_scalar` function, so maybe something like a 3x3 shuffle gadget would be a good example?


hdevalence:
Removed doctest stub and stored the doctest commits in a `circuit-doctests` branch, which we can come back to after #187

Issue #181, opened 2018-10-02
Hero image for the README


oleganza:
This issue is created purely in order to upload an image to GH and serve it from there until we upload it to `https://doc.dalek.rs/assets/bulletproofs-diagram.png`.

<img width="800" alt="bulletproofs-diagram" src="https://user-images.githubusercontent.com/698/46373713-9cc40280-c643-11e8-9bfe-2b0586e40369.png">

oleganza:
See PR #182

PR #182, opened 2018-10-02
Add bulletproofs diagram to the README

oleganza:
This adds a rangeproof construction diagram to the README with background matching the _dark theme_.

PS. Until the image is uploaded to the canonical address at dalek.rs, it is served by Github (see #181).

image


README.md
```diff
@@ -1,5 +1,10 @@
 # Bulletproofs
 
+<img
+ width="100%"
+ src="https://user-images.githubusercontent.com/698/46373713-9cc40280-c643-11e8-9bfe-2b0586e40369.png"
+ TODOsrc="https://doc.dalek.rs/assets/bulletproofs-diagram.png"/>
```
hdevalence: I think this could just use the githubusercontent URL for now, and change the source to doc.dalek.rs later. The reason is that I need to figure out a better system for the `assets` directory for doc.dalek.rs anyways, and some scripts for doing deployment, etc., and that's more work that adding this image doesn't need to be coupled to.

oleganza: Works for me.

oleganza:
PTAL

hdevalence:
https://github.com/dalek-cryptography/dalek_workspace/issues/1

Issue #180, opened 2018-10-01
R1CS benchmark


hdevalence:
We should probably have at least one benchmark.  The benchmarks are totally dependent on the circuit, so if we have a shuffle circuit in the doctests anyways (#179), we might want to use it as a benchmark.  The current benchmarks are in `benches/bulletproofs.rs`; I think this should probably be renamed to `benches/rangeproof.rs` and the R1CS benchmarks can go in `benches/r1cs.rs`.

cathieyun:
PR for shuffle benchmark here: https://github.com/dalek-cryptography/bulletproofs/pull/187

cathieyun:
we merged the kshuffle benchmark into the `circuit` branch, so can we close this issue? Or would we like to add more benchmarks? @hdevalence

PR #187, opened 2018-10-08
Add R1CS benchmark (shuffle)

cathieyun:
Addresses https://github.com/dalek-cryptography/bulletproofs/issues/180


hdevalence:
What's the relation of this branch to #179 ? That PR changes the R1CS API (which affects this code), but I think it was blocked on adding a doctest, although my understanding is that this PR is adding code which we want to turn into the doctest later.

Maybe the best thing to do is to pull the doctest stub out of #179, put this PR on top of #179, then later make another PR that adds a modified version of this code as a doctest?

oleganza:
FYI. I've tried running this with and without a megacheck patch from #159: verification of a 32-item shuffle is 5.1ms before, 1.9ms after. Also, with the megacheck, this is how shuffle gadget compares to a range proof:

```
32-item shuffle:    214 points, 64 mults
64-bit rangeproof:  147 points, 64 mults
```

PR #190, opened 2018-10-11
Efficient padding of the constraint system to a power-of-two multipliers

### LINKED ISSUE #156: Eliminate padding variables
Right now the constraint systems pad to the nearest power of 2 by adding dummy multipliers. But instead we could change the proving and verification logic to pad with zeros, which would reduce the cost of proving and verification.

oleganza:
This adds notes and implementation of a lower-level padding of the proof components as required by the inner product argument.

Closes #156

image

### DIFF
--- a/docs/cs-proof.md
+++ b/docs/cs-proof.md
@@ -562,7 +562,7 @@ S   &= \widetilde{B} \cdot \tilde{s} + \langle \mathbf{G} , \mathbf{s}\_L \rangl
 
 The prover forms these commitments, and sends them to the verifier.
 
-For reference, here are the equations for \\({\mathbf{l}}(x)\\) and transmuted \\({\mathbf{r}}(x)\\):
+For reference, here are the equations for \\({\mathbf{l}}(x)\\), and \\({\mathbf{r}}(x)\\) multiplied by \\(\mathbf{y}^{-n}\\):
 
 \\[
 \begin{aligned}
--- a/docs/cs-proof.md
+++ b/docs/cs-proof.md
@@ -621,32 +621,43 @@ and related commitments before we can use the inner product argument.
 Our goal is to translate the _padding of the constraint system_ into the _padding of proof data_,
 so we can keep the constraint system small and perform less computations in proving and verification.
 
+We will use the following notation for the padding:
+
+\\[
+\begin{aligned}
+                           n^{+} &= 2^{\lceil \log_2 n \rceil} \\\\
+                      {\Delta n} &= n^{+} - n \\\\
+           \mathbf{0}^{\Delta n} &= [0,...,0] \\\\
+  \mathbf{0}^{q \times \Delta n} &= [[0,...,0], ..., [0,...,0]]
+\end{aligned}
+\\]
+
 We start by padding the entire constraint system:
 multipliers are padded with all-zero assignments \\(a\_{L,j}, a\_{R,j}, a\_{O,j}\\),
 all-zero blinding factors \\(s\_{L,j}, s\_{R,j}\\),
 and all-zero weights \\(W\_{R,i,j}, W\_{L,i,j}, W\_{O,i,j}\\),
-for all constraints \\(i \in [0, q)\\) and all additional multipliers \\(j \in [n,n')\\):
+for all constraints \\(i \in [0, q)\\) and all additional multipliers \\(j \in [n,n^{+})\\):
 
 \\[
 \begin{aligned}
-\mathbf{a}\_L^{+} &= \mathbf{a}\_L \hspace{0.1cm} || \hspace{0.1cm} [0,...,0] \\\\
-\mathbf{a}\_R^{+} &= \mathbf{a}\_R \hspace{0.1cm} || \hspace{0.1cm} [0,...,0] \\\\
-\mathbf{a}\_O^{+} &= \mathbf{a}\_O \hspace{0.1cm} || \hspace{0.1cm} [0,...,0] \\\\
-\mathbf{s}\_L^{+} &= \mathbf{s}\_L \hspace{0.1cm} || \hspace{0.1cm} [0,...,0] \\\\
-\mathbf{s}\_R^{+} &= \mathbf{s}\_R \hspace{0.1cm} || \hspace{0.1cm} [0,...,0] \\\\
-\mathbf{W}\_L^{+} &= \mathbf{W}\_L \hspace{0.1cm} || \hspace{0.1cm} [[0,...,0],...,[0,...,0]] \\\\
-\mathbf{W}\_R^{+} &= \mathbf{W}\_R \hspace{0.1cm} || \hspace{0.1cm} [[0,...,0],...,[0,...,0]] \\\\
-\mathbf{W}\_O^{+} &= \mathbf{W}\_O \hspace{0.1cm} || \hspace{0.1cm} [[0,...,0],...,[0,...,0]] \\\\
+\mathbf{a}\_L^{+} &= \mathbf{a}\_L \hspace{0.1cm} || \hspace{0.1cm} \mathbf{0}^{\Delta n} \\\\
+\mathbf{a}\_R^{+} &= \mathbf{a}\_R \hspace{0.1cm} || \hspace{0.1cm} \mathbf{0}^{\Delta n} \\\\
+\mathbf{a}\_O^{+} &= \mathbf{a}\_O \hspace{0.1cm} || \hspace{0.1cm} \mathbf{0}^{\Delta n} \\\\
+\mathbf{s}\_L^{+} &= \mathbf{s}\_L \hspace{0.1cm} || \hspace{0.1cm} \mathbf{0}^{\Delta n} \\\\
+\mathbf{s}\_R^{+} &= \mathbf{s}\_R \hspace{0.1cm} || \hspace{0.1cm} \mathbf{0}^{\Delta n} \\\\
+\mathbf{W}\_L^{+} &= \mathbf{W}\_L \hspace{0.1cm} || \hspace{0.1cm} \mathbf{0}^{q \times \Delta n} \\\\
+\mathbf{W}\_R^{+} &= \mathbf{W}\_R \hspace{0.1cm} || \hspace{0.1cm} \mathbf{0}^{q \times \Delta n} \\\\
+\mathbf{W}\_O^{+} &= \mathbf{W}\_O \hspace{0.1cm} || \hspace{0.1cm} \mathbf{0}^{q \times \Delta n} \\\\
 \end{aligned}
 \\]
 
-As a result, the vectors of generators \\(\mathbf{G},\mathbf{H}\\) and challenges \\(\mathbf{y}^n\\) are extended:
+As a result, we have to take larger slices of the vectors of generators \\(\mathbf{G},\mathbf{H}\\) and more powers of the challenge \\(y\\):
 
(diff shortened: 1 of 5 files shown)


docs/cs-proof.md
```diff
@@ -609,6 +609,115 @@ to prove that
...
+
+We start by padding the entire constraint system:
+multipliers are padded with all-zero assignments \\(a\_{L,j}, a\_{R,j}, a\_{O,j}\\),
+all-zero blinding factors \\(s\_{L,j}, s\_{R,j}\\),
+and all-zero weights \\(W\_{R,i,j}, W\_{L,i,j}, W\_{O,i,j}\\),
+for all constraints \\(i \in [0, q)\\) and all additional multipliers \\(j \in [n,n')\\):
+
+\\[
+\begin{aligned}
+\mathbf{a}\_L^{+} &= \mathbf{a}\_L \hspace{0.1cm} || \hspace{0.1cm} [0,...,0] \\\\
```
hdevalence: Should we write this as `[0, ..., 0]` or as `\mathbf 0`?

docs/cs-proof.md
```diff
@@ -609,6 +609,115 @@ to prove that
...
+\mathbf{a}\_O^{+} &= \mathbf{a}\_O \hspace{0.1cm} || \hspace{0.1cm} [0,...,0] \\\\
+\mathbf{s}\_L^{+} &= \mathbf{s}\_L \hspace{0.1cm} || \hspace{0.1cm} [0,...,0] \\\\
+\mathbf{s}\_R^{+} &= \mathbf{s}\_R \hspace{0.1cm} || \hspace{0.1cm} [0,...,0] \\\\
+\mathbf{W}\_L^{+} &= \mathbf{W}\_L \hspace{0.1cm} || \hspace{0.1cm} [[0,...,0],...,[0,...,0]] \\\\
+\mathbf{W}\_R^{+} &= \mathbf{W}\_R \hspace{0.1cm} || \hspace{0.1cm} [[0,...,0],...,[0,...,0]] \\\\
+\mathbf{W}\_O^{+} &= \mathbf{W}\_O \hspace{0.1cm} || \hspace{0.1cm} [[0,...,0],...,[0,...,0]] \\\\
+\end{aligned}
+\\]
+
+As a result, the vectors of generators \\(\mathbf{G},\mathbf{H}\\) and challenges \\(\mathbf{y}^n\\) are extended:
```
hdevalence: This seems like it might be a good place to mention how we deal with the generators -- we're pulling G and H from an arbitrarily-long chain of XOF outputs, so there's a natural way to extend the G and H, and it might be cleaner to mention this, so that we could just say that G+ and H+ are just bigger sections, or something

docs/cs-proof.md
```diff
@@ -609,6 +609,115 @@ to prove that
...
+\\]
+
+The low-level variables are padded with zeroes, so their commitments remain unchanged:
+
+\\[
+\begin{aligned}
+A_I^{+} &= \widetilde{B} \cdot \tilde{a} + \langle \mathbf{G}^{+} , \mathbf{a}\_L^{+} \rangle + \langle \mathbf{H}^{+}, \mathbf{a}\_R^{+} \rangle &{}={}& A_I \\\\
+A_O^{+} &= \widetilde{B} \cdot \tilde{o} + \langle \mathbf{G}^{+} , \mathbf{a}\_O^{+} \rangle                                                     &{}={}& A_O \\\\
+S^{+}   &= \widetilde{B} \cdot \tilde{s} + \langle \mathbf{G}^{+} , \mathbf{s}\_L^{+} \rangle + \langle \mathbf{H}^{+}, \mathbf{s}\_R^{+} \rangle &{}={}& S
+\end{aligned}
```
hdevalence: Might be slightly more clean if this shows the step of breaking the a_L+ into two parts, one of which is all zeros

docs/cs-proof.md
```diff
@@ -609,6 +609,115 @@ to prove that
...
+\\]
+
+Vector polynomials \\(\mathbf{l}(x)\\) is padded with zeroes and \\(\mathbf{r}(x)\\) is padded with additional powers of \\(y\\) in its 0th term:
+
+\\[
+\begin{aligned}
+\mathbf{l}(x)^{+} &= \mathbf{a}\_L^{+} \cdot x + \mathbf{s}\_L^{+} \cdot x^3 + \mathbf{y}^{-n^{+}} \circ \mathbf{w}\_R^{+} \cdot x + \mathbf{a}\_O^{+} \cdot x^2 \\\\
+                  &= \mathbf{a}\_L \cdot x + \mathbf{s}\_L \cdot x^3 + \mathbf{y}^{-n} \circ \mathbf{w}\_R \cdot x + \mathbf{a}\_O \cdot x^2 \\\\
+                  &  \hspace{0.5cm} || \hspace{0.1cm} [0,...,0] \cdot x + [0,...,0] \cdot x^3 + [y^n,...,y^{n^{+}-1}] \circ [0,...,0] \cdot x + [0,...,0] \cdot x^2 \\\\
+                  &= \mathbf{l}(x) || [0,...,0] \\\\
```
hdevalence: This might be good to split into two blocks, one for l and one for r

docs/cs-proof.md
```diff
@@ -609,6 +609,115 @@ to prove that
...
+\\[
+\begin{aligned}
+\delta(y, z)^{+} &= \langle \mathbf{y}^{-n^{+}} \circ \mathbf{w}\_R^{+}, \mathbf{w}\_L^{+} \rangle \\\\
+                 &= \langle \mathbf{y}^{-n} \circ \mathbf{w}\_R, \mathbf{w}\_L \rangle      +     \langle [y^n,...,y^{n^{+}-1}] \circ [0,...,0], [0,...,0] \rangle \\\\
+                 &= \langle \mathbf{y}^{-n} \circ \mathbf{w}\_R, \mathbf{w}\_L \rangle      +     0 \\\\
+                 &= \delta(y, z)
+\end{aligned}
+\\]
+
+Vector polynomials \\(\mathbf{l}(x)\\) is padded with zeroes and \\(\mathbf{r}(x)\\) is padded with additional powers of \\(y\\) in its 0th term:
```
hdevalence: > \\(\mathbf{r}(x)\\) is padded with additional powers of \\(y\\) in its 0th term:

This was pretty unclear when I read it, maybe it would be good to move it right before the `r` calculation and just say that it's padded with additional powers of y? To me, saying "in its 0th term" makes it sound like the change is in the first part of the vector, rather than saying that the padding does not have any x coefficients.

docs/cs-proof.md
```diff
@@ -752,10 +874,21 @@ The verifier flattens constraints:
 \\]
 where each of \\(\mathbf{w}\_L, \mathbf{w}\_R, \mathbf{w}\_O\\) has length \\(n\\) and \\(\mathbf{w}\_V\\) has length \\(m\\).
 
+The verifier [pads](#padding-mathbflx-and-mathbfrx-for-the-inner-product-proof) the generators \\(\mathbf{G},\mathbf{H}\\) and challenges \\(\mathbf{y}^n\\) to a power-of-two \\(n^{+}\\):
```
hdevalence: This also talks about the generator-extension as padding, but I'm not sure if it's exactly the right metaphor: when we do proving and verification, the generators are already given up front with some capacity, and we use the first power-of-two subset of them.  So rather than padding the generators, it's selecting a bigger subslice of a longer set of generators.

docs/cs-proof.md
cathieyun: maybe \mathbf 0^{n^+ - n} (or whatever the length of the zero vector is for that instance) would be the most informative... right now, it's not super clear what its length is

oleganza: Yeah, i’ll introduce a `delta n` probably to make it all concrete.

hdevalence: Does it matter what its length is?

cathieyun: in some cases yes - such as for `l(x)+`, it would be nice to know how many zeroes the vector is padded with. (you can deduce from following the math or comparing it to `r(x)+`, but it seems nice to note the length explicitly)

oleganza:
@hdevalence thanks for the feedback. I've made it all more precise in ac9cd54.

docs/cs-proof.md
hdevalence: Hmm, I'm not sure I understand why you would need to know how many zeros the vector is padded with, and if you don't, it seems better not to give it a notation at all (to reduce cognitive overhead and hint that it's not important).

oleganza: Looks cleaner indeed:

image

docs/cs-proof.md
oleganza: Done in 72fae27

oleganza:
Implemented padding in the prover and verifier, and added a k-shuffle to the tests.

oleganza:
(rebased on the latest `circuit`)

docs/cs-proof.md
```diff
@@ -556,18 +556,18 @@ A_O &= \widetilde{B} \cdot \tilde{o} + \langle \mathbf{G} , \mathbf{a}\_O \rangl
...
 W_O &= \langle \mathbf{y}^{-n} \circ \mathbf{w}\_O, \mathbf{H} \rangle \\\\
-S &= \widetilde{B} \cdot \tilde{s} + \langle \mathbf{G} , \mathbf{s}\_L \rangle + \langle \mathbf{H}, \mathbf{s}\_R \rangle
+S   &= \widetilde{B} \cdot \tilde{s} + \langle \mathbf{G} , \mathbf{s}\_L \rangle + \langle \mathbf{H}, \mathbf{s}\_R \rangle
 \end{aligned}
 \\]
 
 The prover forms these commitments, and sends them to the verifier.
 
-For reference, here are the equations for \\({\mathbf{l}}(x)\\) and \\({\mathbf{r}}(x)\\):
+For reference, here are the equations for \\({\mathbf{l}}(x)\\) and transmuted \\({\mathbf{r}}(x)\\):
```
hdevalence: In what sense is `r` transmuted?

src/circuit_proof/prover.rs
```diff
@@ -366,8 +358,22 @@ impl<'a, 'b> ProverCS<'a, 'b> {
 
         let t_x = t_poly.eval(x);
         let t_x_blinding = t_blinding_poly.eval(x);
-        let l_vec = l_poly.eval(x);
-        let r_vec = r_poly.eval(x);
+        let l_vec = l_poly
+            .eval(x)
+            .into_iter()
+            .chain(iter::repeat(Scalar::zero()).take(pad))
+            .collect::<Vec<_>>();
```
hdevalence: might be cleaner as `l_poly.eval(x).append(vec![Scalar::zero(); pad]);`

docs/cs-proof.md
oleganza: Ehh... i wanted to move `y^n` on the other side of the equation to match with the following 2d diagram. 

Maybe I should remove "transmuted" and add ", where the factor `y^n` is moved to the left side of the equation"

src/circuit_proof/prover.rs
oleganza: thanks, good point

docs/cs-proof.md
oleganza: Changed to 
> For reference, here are the equations for `l(x)`, and `r(x)` multiplied by `y^-n`

image

PR #193, opened 2018-10-16
Zero heap allocations in the circuit code

### LINKED ISSUE #146: Constraint system code should zero its heap allocations.
This probably isn't worth doing until we're mostly finished writing the constraint system code, but we should do it before we finish it.

oleganza:
Closes #146

### DIFF
--- a/src/circuit_proof/assignment.rs
+++ b/src/circuit_proof/assignment.rs
@@ -20,7 +20,9 @@ pub enum Assignment {
 
 // Default implementation is used for zeroizing secrets from allocated memory via `clear_on_drop`.
 impl Default for Assignment {
-    fn default() -> Assignment { Assignment::Missing() }
+    fn default() -> Assignment {
+        Assignment::Missing()
+    }
 }
 
 impl From<Option<Scalar>> for Assignment {
--- a/src/circuit_proof/assignment.rs
+++ b/src/circuit_proof/assignment.rs
@@ -18,6 +18,11 @@ pub enum Assignment {
     Missing(),
 }
 
+// Default implementation is used for zeroizing secrets from allocated memory via `clear_on_drop`.
+impl Default for Assignment {
+    fn default() -> Assignment { Assignment::Missing() }
+}
+
 impl From<Option<Scalar>> for Assignment {
     fn from(o: Option<Scalar>) -> Self {
         match o {
--- a/src/circuit_proof/assignment.rs
+++ b/src/circuit_proof/assignment.rs
@@ -18,7 +18,7 @@ pub enum Assignment {
     Missing(),
 }
 
-// Default implementation is used for zeroizing secrets from allocated memory via `clear_on_drop`.
+// Default implementation is used for zeroing secrets from allocated memory via `clear_on_drop`.
 impl Default for Assignment {
     fn default() -> Assignment {
         Assignment::Missing()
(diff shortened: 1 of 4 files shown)


src/circuit_proof/assignment.rs
```diff
@@ -18,6 +18,13 @@ pub enum Assignment {
     Missing(),
 }
 
+// Default implementation is used for zeroizing secrets from allocated memory via `clear_on_drop`.
```
cathieyun: zeroizing -> zeroing?

src/circuit_proof/prover.rs
```diff
@@ -389,6 +413,16 @@ impl<'a, 'b> ProverCS<'a, 'b> {
...
+        // We do not yet have a ClearOnDrop wrapper for Vec<Scalar>.
+        // When PR 202 [1] is merged, we can simply wrap s_L and s_R at the point of creation.
+        // [1] https://github.com/dalek-cryptography/curve25519-dalek/pull/202
+        for e in s_L.iter_mut() {
+            e.clear();
+        }
+        for e in s_R.iter_mut() {
+            e.clear();
+        }
+
```
cathieyun: do we also have to clear i_blinding, o_blinding, and s_blinding?

src/circuit_proof/assignment.rs
oleganza: lol, what did i write?

src/circuit_proof/prover.rs
oleganza: Maybe... I've [added](
https://github.com/dalek-cryptography/bulletproofs/pull/193/commits/da06f2c2b85b8444ca86bfce951848447686b036) clearing for similar stack variables in the rangeproof code that i overlooked in the previous PR #183.

@hdevalence wdyt, should the stack allocations be cleared too?

hdevalence: I think we shouldn't include `da06f2c2b85b` (I guess I'm a minute too late), since it doesn't seem like the right way to clear stack allocations.

hdevalence: Manually clearing a few particular stack variables isn't really effective, because what you need to clear is both the secrets as well as all secret-derived data.  For heap allocations, the idea is that the secret-derived data is probably going to live on the stack, so the only zeroing targets on the heap are the variables themselves.  But if you want to try to clear stack data, you really need to wipe the entire stack.  This is difficult because as the programmer, you don't actually control the stack allocation.

More specifically, the commit that adds the clearing for stack variables makes mutable a bunch of data that should be immutable, but also doesn't actually erase all the stack data, so it doesn't seem like an improvement.

cathieyun: @hdevalence sorry for not giving you time to respond before the LGTM / merge. What you said about stack variables makes sense. @oleganza wdyt of opening a new PR to revert the clearing of stack variables?

oleganza: >More specifically, the commit that adds the clearing for stack variables makes mutable a bunch of data that should be immutable, but also doesn't actually erase all the stack data, so it doesn't seem like an improvement.

Yup. Will add a reverting PR in a sec.

oleganza: @cathieyun that's my bad - i asked a question and just merged away before receiving an answer.

oleganza: @hdevalence @cathieyun PTAL? https://github.com/dalek-cryptography/bulletproofs/pull/195

Issue #186, opened 2018-10-06
Two-phase circuit construction with intermediate commitments


oleganza:
### Problem

Our current R1CS interface allows constructing constraints using challenges bound to the external variables (**V**). This, however, may be risky for complex circuits. For example, a _shuffle gadget_ is only binding if inputs on both sides are committed. However, if one side remains uncommitted (e.g. when there is a circuit in between two shuffles and only "outer" sides are ultimately committed), then the low-level variables on that side can be forged based on the knowledge of the challenge.

### Disclaimer

This is an over-the-napkin draft of a maybe-solution. Little attempt was made to analyze the alternatives or soundness of this proposal.

### Goal

Make gadgets _locally sound_ by giving an opportunity to commit to input low-level variables, regardless of whether they are indirectly committed via high-level variables.

### Proposal

We split the circuit construction in two phases: _wireframing phase_ and _completion phase_.

**Wireframing phase**

Gadgets can allocate low-level variables, add constraints, but cannot use challenges. This phase produces first pair of vector commitments `A_I1` and `A_O1` to these variables. The constraint system trait is called `WireframeCS`.

This allow allocating all the gadgets and all variables between them, assigning required values to these variables. But not all gadgets can implement their internal constraints yet because they need challenge values.

The phase ends with two points `A_I1` and `A_O1` computed and sent to the transcript, committing to all the low-level variables allocated so far.

**Completion phase**

Gadgets can allocate more low-level variables, add more constraints and can use challenges. This phase produces second pair of vector commitments `A_I2` and `A_O2` to these additional low-level variables. The constraint system trait is called `CompleteCS`.

This allows implementing gadgets, such as _shuffle_ or _merge_ using challenges, making them **locally sound** because all input variables for any given gadget are already committed via `A_I1/A_O1`.

After the Completion Phase is complete, full commitments to the low level variables are computed and sent to the transcript: 

```
transcript.commit_point(A_I2);
transcript.commit_point(A_O2);
let e = transcript.challenge_scalar();
let A_I = A_I1 + e*A_I2;
let A_O = A_O1 + e*A_O2;
```

The verifier's logic is:

1. The proof now contains 2 more points (`A_I1,A_I2,A_O1,A_O2` instead of `A_I,A_O`).
2. The verifier starts by making the _wireframe_ of the circuit: allocating variables and adding constraints _that do not use challenge values_.
3. Then, the verifier sends `A_I1, A_O1` to the transcript.
[...]

PR #196, opened 2018-10-18
[WIP] Two-phase constraint system with intermediate commitments

oleganza:
This addresses #186. We break `A_{I,O}` into two commitments and prove two statements about two parts of the vectors `l(x), r(x)` independently to make sure the commitments are well-formed. Then, we merge those two statements in one using a challenge. Then, we transmute the commitments of the second subslices of the vectors affected by the challenge to commitments to these vectors without the challenge, but on challenge-tweaked generators.


oleganza:
Closing this proof-of-concept in order to discuss first in #222 and be certain about the design and rationale.

PR #200, opened 2018-11-06
Some note touchups [WIP]

cathieyun:
I realized that there are some parts in the write-up that have a gap in explanation / require a big jump in understanding, and I’d like to fill those in a little more.

Also, fix up the rangeproof doctests & add an example doctest to the README.


oleganza:
FYI: I made some edits on the aggregation in the `circuit` branch. In case you want to include them here. Cf. https://github.com/dalek-cryptography/bulletproofs/pull/191/files

cathieyun:
I based the branch off of `circuit` instead of `main`, whoops.
I created a new PR that is based off of `main`... see here:
https://github.com/dalek-cryptography/bulletproofs/pull/201

PR #202, opened 2018-11-06
Explicitly cap allocations inside IPP verification

oleganza:
This patch eliminates blind allocation of `O(2^k)` array inside the inner product verification function based on the untrusted input of `O(k)` length.

For the rangeproof the cap is enforced in the rangeproof layer that caps the `2^k` to `64*<parties>` and for the constraint system it could be explicitly written as a metadata on a proof.

### DIFF
--- a/src/inner_product_proof.rs
+++ b/src/inner_product_proof.rs
@@ -184,7 +184,7 @@ impl InnerProductProof {
         transcript: &mut Transcript,
     ) -> Result<(Vec<Scalar>, Vec<Scalar>, Vec<Scalar>), ProofError> {
         let lg_n = self.L_vec.len();
-        if lg_n > 32 {
+        if lg_n >= 32 {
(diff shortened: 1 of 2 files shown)


src/inner_product_proof.rs
```diff
@@ -177,12 +177,21 @@ impl InnerProductProof {
 
     /// Computes three vectors of verification scalars \\([u\_{i}^{2}]\\), \\([u\_{i}^{-2}]\\) and \\([s\_{i}]\\) for combined multiscalar multiplication
     /// in a parent protocol. See [inner product protocol notes](index.html#verification-equation) for details.
+    /// Verifier must provide the input length \\(n\\) explicitly to avoid unbounded allocation within the inner product proof.
```
hdevalence:
```suggestion
    /// The verifier must provide the input length \\(n\\) explicitly to avoid unbounded allocation within the inner product proof.
```

PR #203, opened 2018-11-06
Versions for RP and IPP domain separators + cleaner test

oleganza:


src/transcript.rs
```diff
@@ -26,13 +26,13 @@ fn le_u64(value: u64) -> [u8; 8] {
 
 impl TranscriptProtocol for Transcript {
     fn rangeproof_domain_sep(&mut self, n: u64, m: u64) {
-        self.commit_bytes(b"dom-sep", b"rangeproof");
+        self.commit_bytes(b"dom-sep", b"rangeproof1");
```
hdevalence:
```suggestion
        self.commit_bytes(b"dom-sep", b"rangeproof v1");
```

src/transcript.rs
```diff
@@ -26,13 +26,13 @@ fn le_u64(value: u64) -> [u8; 8] {
...
         self.commit_bytes(b"n", &le_u64(n));
         self.commit_bytes(b"m", &le_u64(m));
     }
 
     fn innerproduct_domain_sep(&mut self, n: u64) {
-        self.commit_bytes(b"dom-sep", b"ipp");
+        self.commit_bytes(b"dom-sep", b"ipp1");
```
hdevalence:
```suggestion
        self.commit_bytes(b"dom-sep", b"ipp v1");
```

src/transcript.rs
hdevalence: (sorry for bikeshedding)

PR #207, opened 2018-11-08
Merge main into circuit

hdevalence:
This causes a test failure on the zero-length inner products (due to the changed logic in length checking for verification).


hdevalence:
Updated to remove the 0-length IPP, which is only used for making constraint systems with no multipliers; those proofs could be better done using Schnorr proofs.

src/inner_product_proof.rs
```diff
@@ -51,15 +51,6 @@ impl InnerProductProof {
...
-        if n == 0 {
-            return InnerProductProof {
-                L_vec: vec![],
-                R_vec: vec![],
-                a: Scalar::zero(),
-                b: Scalar::zero(),
-            };
-        }
-
         // All of the input vectors must have the same length.
```
oleganza: Seems like `if n == 0` check got lost in this merge

oleganza: Oh, I misunderstood your comment. I thought that's the check that you've added.

Why don't we make the API complete and allow zero multipliers, but improve the check in `verification_scalars` to handle zero correctly? Even if no-multiplier case is better handled by schnorr proofs, the user of a higher-level protocol may not even be exposed to the amount of multipliers and can accidentally create constraints without any.

hdevalence: Any case where you would make a size-0 IPP you could just prove nothing, so I think that rejecting 0 is fine.

oleganza: In a R1CS, internal 0-IPP would mean verifying linear constraints on a bunch of commitments, so it's implementing basically a schnorr proof, just with a few unnecessary challenges. So while IPP part proves nothing, CS proof is still useful.

oleganza: I mean, 0-multiplier r1cs proof is proving a statement `<W_v, V> + W_c = 0` which is not nothing.

hdevalence:
Would it be possible to not include the "R1CS" -> "ConstraintSystem" change in this PR? I didn't see it in #210, but I think that "R1CS" is a better name because it's more precise, it's a standard terminology, and it's shorter.

cathieyun:
👀

PR #210, opened 2018-11-13
Small edits to the circuit code

cathieyun:
cleanups:
- doc updates
- fixme notes
- update benchmark, doctest code
- rename funcs
etc etc


src/util.rs
```diff
@@ -112,21 +112,19 @@ impl VecPoly3 {
...
-    // MUST call l_poly.inner_product(r_poly).
-    pub fn inner_product(&self, rhs: &VecPoly3) -> Poly6 {
-        let l = self;
-        let r = rhs;
-
+    /// Compute an inner product of `lhs`, `rhs` which have the property that:
+    /// - `lhs.0` is zero;
+    /// - `rhs.2` is zero;
+    /// This is the case in the constraint system proof.
+    pub fn special_inner_product(lhs: &Self, rhs: &Self) -> Poly6 {
```
oleganza: 👍 for reflecting this in the fn name.

benches/r1cs.rs
```diff
@@ -69,128 +69,111 @@ For K = 1:
...
+/// The inputs and outputs are all tuples of the `Variable, Assignment`, where the `Assignment`
+/// can be either assigned as `Value::Scalar` or unassigned as `Missing`.
+pub fn fill_cs<CS: ConstraintSystem>(
+    cs: &mut CS,
+    x: Vec<(Variable, Assignment)>,
+    y: Vec<(Variable, Assignment)>,
+) -> Result<(), KShuffleError> {
+    let one = Scalar::one();
+    let z = cs.challenge_scalar(b"k-scalar shuffle challenge");
+    let neg_z = -z;
```
oleganza: nitpick: we can make the code a bit cleaner by just passing `z` and subtracting it when needed - the difference in performance of `Add` and `Sub` is really insignificant if there is any.

src/lib.rs
```diff
@@ -29,7 +29,7 @@ mod util;
 
 #[doc(include = "../docs/notes.md")]
 mod notes {}
-mod circuit_proof;
+mod constraint_system;
```
oleganza: (Maybe in the separate PR or even later) My bikeshed preference for names is: 

1. Rename private module `constraint_system` into `cs`.
2. Rename public module with reexports from `r1cs` into `cs`. All crate-private identifiers could be made `pub(crate)` or we can do `use cs as cs_internal` inside `mod.rs` so we avoid using `pub(crate)` and can create another module `cs` for public re-exports.
3. Rename `R1CSProof` into `ConstraintSystemProof`.

oleganza: I'm also down with `s/cs/constraint_system/` if you prefer the spelled out name. But for both private and public modules.

oleganza: How about internal “cs” and external “constraint_system”?

cathieyun:
addressed suggestions, will merge after the CI tests pass, unless there is more feedback!

hdevalence:
LGTM

src/lib.rs
hdevalence: I think that "R1CS" is a good name and that we should keep it.

PR #213, opened 2018-11-14
Circuit trait rework

cathieyun:
When going through the docs, we realized that it's possible to make the API simpler and more elegant - instead of allocating variables and then adding them to the constraint system in linear constraints, we could do both of those steps at the same time by just passing linear constraints directly to the `add_constraint()` function. Then the constraint system can internally evaluate the linear constraints, meaning the user doesn't have to pass in `Assignment`s.


hdevalence:
screenshot from 2018-11-14 11-13-22
:heart_eyes_cat:

src/r1cs/tests.rs
```diff
@@ -176,132 +160,43 @@ For K = 1:
...
-    }
-}
-
 impl KShuffleGadget {
-    pub fn fill_cs<CS: ConstraintSystem>(
-        cs: &mut CS,
-        x: Vec<(Variable, Assignment)>,
-        y: Vec<(Variable, Assignment)>,
-    ) -> Result<(), KShuffleError> {
+    pub fn fill_cs<CS: ConstraintSystem>(cs: &mut CS, x: Vec<Variable>, y: Vec<Variable>) {
```
hdevalence: Another thought on the shuffle example: since the shuffle uses a challenge that's bound only to the HL witness variables, it's not really safely composable with other gadgets, so perhaps we should change the code to be a `KShuffleProof` with `prove`, `verify` methods and an internal `gadget` function? This way we would have an opportunity to demonstrate not just making a gadget function, but also how to encapsulate the ProverCS / VerifierCS setup, and to talk about what the challenges are bound to and why it matters.

cathieyun: 👍

src/r1cs/tests.rs
```diff
@@ -176,289 +160,189 @@ For K = 1:
...
-
-        // Make multiplier gate variables
-        let (mul_left_var, mul_right_var, mul_out_var) =
-            cs.assign_multiplier(mul_left, mul_right, mul_out)?;
+    pub fn prove<'a, 'b>(
+        pc_gens: &'b PedersenGens,
+        bp_gens: &'b BulletproofGens,
+        transcript: &'a mut Transcript,
+        input: &Vec<u64>,
+        output: &Vec<u64>,
```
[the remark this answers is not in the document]
hdevalence: Ah, good point!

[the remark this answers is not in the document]
cathieyun: Fixed, input and out are `&[Scalar]` now!

benches/r1cs.rs
```diff
@@ -72,181 +69,113 @@ For K = 1:
...
-        intermediate_multiplier(cs, z, prev_out?, x[i])
-    })?;
+// Make a gadget that adds constraints to a ConstraintSystem, such that the
+// y variables are constrained to be a valid shuffle of the x variables.
+struct KShuffleGadget {}
 
-    // Make last y multiplier for i = k-1 and k-2
-    let last_muly_out = last_multiplier(cs, z, y[k - 1], y[k - 2]);
+impl KShuffleGadget {
+    fn fill_cs<CS: ConstraintSystem>(cs: &mut CS, x: Vec<Variable>, y: Vec<Variable>) {
```
hdevalence: Is there a reason that this needs to take ownership of the `Variable`s, or could it be
```suggestion
    fn shuffle_gadget<CS: ConstraintSystem>(cs: &mut CS, x: &[Variable], y: &[Variable]) {
```
Then the code that calls the gadget in the prove and verify functions can save an alloc by using `split_at` on the `Vec<Variable>` returned from the constraint system

hdevalence: Oops, I should have put this comment in the version that's in the test code, not the version that's in the benchmark

cathieyun: No reason! That looks good, will update

PR #219, opened 2018-11-16
Edit doctest example [WIP]

hdevalence:
Reformat the shuffle proof doctest in the r1cs docs


docs/r1cs-docs-example.md
```diff
@@ -0,0 +1,530 @@
...
+
+        (x_0 - z)---⊗------⊗---(y_0 - z)        // mulx[0], muly[0]
+                    |      |
+        (x_1 - z)---⊗      ⊗---(y_1 - z)        // mulx[1], muly[1]
+                    |      |
+                   ...    ...
+                    |      |
+    (x_{k-2} - z)---⊗      ⊗---(y_{k-2} - z)    // mulx[k-2], muly[k-2]
+                   /        \
+    (x_{k-1} - z)_/          \_(y_{k-1} - z)
```
hdevalence: We could keep this diagram, but the rest of the block should probably be doc comments?

docs/r1cs-docs-example.md
```diff
@@ -0,0 +1,530 @@
+The rank-1 constraint system API for programmatically defining constraint systems.
+
+## Building a proof-of-shuffle constraint system
+
+```ascii,no_run
+A shuffle is a permutation of a list of `k` scalars `{x_i}` into a list of `k` scalars `{y_i}`.
+
+Algebraically it can be expressed as a statement that for a free variable `z`, the roots of the two polynomials in terms of `z` are the same up to a permutation:
+
+    ∏(x_i - z) == ∏(y_i - z)
```
hdevalence:
```suggestion
\\[
\prod_i (x_i - z) = \prod_i (y_i - z)
\\]
```

docs/r1cs-docs-example.md
```diff
@@ -0,0 +1,530 @@
...
+	    muly_left[k-2]  = y_{k-2} - z
+	    muly_right[k-2] = y_{k-1} - z
+
+For K = 1:
+	Connect x to y directly, omitting the challenge entirely as it cancels out
+    	x_0 = y_0
+```
+
+This shuffle `prove` function takes a list of scalar inputs and outputs, makes commitments to them, and creates a proof that the committed outputs are a valid reordering of the committed inputs. 
+The shuffle `verify` function verifies that, given a shuffle proof and a list of committed inputs and outputs, the outputs are a valid reordering of the inputs.
```
hdevalence: Since we haven't put the prove and verify functions in the example yet, these should maybe be moved to the prove/venify sections respectively?

docs/r1cs-docs-example.md
```diff
@@ -0,0 +1,530 @@
...
+
+Because only the prover knows the scalar values of the inputs and outputs, and the verifier only sees the (blinded) committed inputs and outputs, the verifier has no knowledge of what the underlying inputs and output values are. Therefore, the only information the verifier learns from this protocol is whether or not the committed outputs are a valid shuffle of the committed inputs - this is why it is a zero-knowledge proof.
```
hdevalence: "(blinded) committed" seems redundant, maybe "only sees the commitments to the inputs ... not the values themselves" would give the same emphasis on the fact that they're hidden?

docs/r1cs-docs-example.md
```diff
@@ -0,0 +1,530 @@
...
+
+First, the function gets a challenge scalar `z` by calling the `ConstraintSystem::challenge_scalar`. This challenge is generated from commitments to high-level variables that were passed to the `ConstraintSystem` when it was created. As noted in the `challenge_scalar` documentation, it is important to ensure that the challenge is sound by making sure that low-level variables are bound to the high-level variables. 
+
+After a check for the lengths of `x` and `y`, the function then makes multipliers to create polynomials in terms of the challenge scalar `z`. It starts with the last multipliers, representing \\( (x_{k-1} - z) * (x_{k-2} - z) \\) and \\( (y_{k-1} - z) * (y_{k-2} - z) \\). The outputs to these last multipliers than become an input to the next multiplier. This continues recursively until it reaches \\( x_0 \\) and \\(y_0\\). Then, it adds a constraint that \\( x_0 = y_0 \\), which constrains that the two polynomials in terms of challenge scalar `z` are equal to each other. This is true if and only if `y` is a valid reordering of `x`.
```
hdevalence: > adds a constraint that \\( x_0 = y_0 \\)

the constraint is on the final products, not the x_0 & y_0, right?

docs/r1cs-docs-example.md
```diff
@@ -0,0 +1,530 @@
...
+        let proof = cs.prove()?;
+
+        Ok((ShuffleProof(proof), commitments))
+    }
+}
+```
+
+## Verifiying a proof
+
+The verifier receives a proof, and a list of committed inputs and outputs, from the prover. It passes these to the `verify` function, and receives a `Result::ok()` if the proof verified correctly and a `Result::error(R1CSError)` otherwise. The proof could fail to verify for a variety of reasons - if the original scalar outputs were not a valid reordering of the inputs, if the commitments were generated or transmitted incorrectly, or if the constraint system was built differently for the prover and verifier.
```
hdevalence: It might be simpler just to say that the proof will verify, or not, since the user shouldn't have to know all the reasons why the proof could fail (and we don't expose them in the error anyways, on purpose)

docs/r1cs-docs-example.md
```diff
@@ -0,0 +1,530 @@
...
+        cs.verify(&self.0)
+    }
+}
+```
+
+## Using the `ShuffleProof`
+
+Here, we use the `ShuffleProof` gadget by first constructing the common inputs to the `prove` and `verify` functions: the Pedersen and Bulletproofs generators. Next, the prover makes the other inputs to the `prove` function: the list of scalar inputs, the list of scalar outputs, and the prover transcript. The prover calls the `prove` function, and gets a proof and a list of committed points that represent the commitments to the input and output scalars.
+
+The prover passes the proof and the commitments to the verifier. The verifier then makes the other inputs to the `verify` function: the verifier transcript. Note that the starting transcript state must be the same for the prover and verifer transcript; if they are not, then the prover and verifier will receive different challenge scalars and the proof will not verify correctly. The verifier then calls the `verify` function, and gets back a `Result` representing whether or not the proof verified correctly.
```
hdevalence: we could note that the starting transcript state is what provides domain separation?

docs/r1cs-docs-example.md
cathieyun: 👍

docs/r1cs-docs-example.md
```diff
@@ -0,0 +1,530 @@
...
+
+        // Constrain last x mul output and last y mul output to be equal
+        cs.constrain(first_mulx_out - first_muly_out);
+    }
+}
+```
+
+In this example, `ShuffleProof::gadget()` is private function that adds constraints to the constraint system that enforce that `y` (the outputs) are a valid reordering of `x` (the inputs). 
+
+First, the function gets a challenge scalar `z` by calling the `ConstraintSystem::challenge_scalar`. This challenge is generated from commitments to high-level variables that were passed to the `ConstraintSystem` when it was created. As noted in the `challenge_scalar` documentation, it is important to ensure that the challenge is sound by making sure that low-level variables are bound to the high-level variables.
```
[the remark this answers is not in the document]
cathieyun: that makes sense. changed to:

As noted in the `challenge_scalar` documentation, making sure that the challenge circuit is sound requires analysis. In this example, the challenge circuit is sound because the challenge is bound to all of the shuffle inputs and outputs, since the inputs and outputs are high-level variables.

docs/r1cs-docs-example.md
cathieyun: whoops, good catch

docs/r1cs-docs-example.md
```diff
@@ -25,31 +29,26 @@ For K > 1:
...
 For K = 1:
-	Connect x to y directly, omitting the challenge entirely as it cancels out
-    	x_0 = y_0
-```
-
-This shuffle `prove` function takes a list of scalar inputs and outputs, makes commitments to them, and creates a proof that the committed outputs are a valid reordering of the committed inputs. 
-The shuffle `verify` function verifies that, given a shuffle proof and a list of committed inputs and outputs, the outputs are a valid reordering of the inputs.
-
-Because only the prover knows the scalar values of the inputs and outputs, and the verifier only sees the (blinded) committed inputs and outputs, the verifier has no knowledge of what the underlying inputs and output values are. Therefore, the only information the verifier learns from this protocol is whether or not the committed outputs are a valid shuffle of the committed inputs - this is why it is a zero-knowledge proof.
+Connect x to y directly, omitting the challenge entirely as it cancels out
```
hdevalence: It might be good to note that there's only one permutation of a 1-element list (so it makes sense that the challenges cancel out)

docs/r1cs-docs-example.md
```diff
@@ -0,0 +1,541 @@
+The rank-1 constraint system API for programmatically defining constraint systems.
+
+## Building a proof-of-shuffle constraint system
+
+A shuffle is a permutation of a list of `k` scalars `{x_i}` into a list of `k` scalars `{y_i}`.
```
hdevalence:
```suggestion
A shuffle is a permutation of a list of \\(k\\) scalars \\(x_i\\) into a list of \\(k\\) scalars \\(y_i\\).
```
change from code block to katex

docs/r1cs-docs-example.md
```diff
@@ -0,0 +1,541 @@
...
+
+Algebraically it can be expressed as a statement that for a free variable `z`, the roots of the two polynomials in terms of `z` are the same up to a permutation:
```
hdevalence:
```suggestion
Algebraically it can be expressed as a statement that for a free variable \\(z\\), the roots of the two polynomials in terms of \\(z\\) are the same up to a permutation:
```
katex

Issue #218, opened 2018-11-16
More ergonomic API to avoid flattening/unflattening of high-level variables


oleganza:
**Problem**

If you have structured high-level variables (e.g. lists of tuples, like inputs and outputs of confidential values for a shuffle gadget), then it's pretty annoying to flatten them in one list of scalars, which is turned into commitments and variables, and then variables are used to construct back the structured set of "Allocated Values" together with putting back their assignments from the original structured input.

**Proposal**

Instead of passing a list of all scalars and a list of their blinding factors, allow individual commitment of them:

```
prover.commit_external_variable(v,v_blinding) -> (RistrettoPoint, Variable)
``` 

which internally creates a commitment, writes it to the transcript, allocates variable and returns the commitment and a variable.

Then, the user can traverse their structured set of secrets and commit them one by one, creating allocated+assigned types (specific to their protocol) along the way. 

E.g. if the user has a list `Vec<Value>`, it can be `map`'d to `Vec<AllocatedValue>` much more easily than what we have to do today.

cathieyun:
This sounds helpful!

hdevalence:
Hmm, right now the external commitments are passed into the constraint system constructor in order to ensure that it's not possible to start building constraints without having declared all of the inputs to the constraint system.  Wouldn't this function break that, unless it was on a `Builder` of some kind?

cathieyun:
I see the concern - if this were implemented there would have to be two "states", one where the user is still adding high-level variables, and one where the user is done and wants to allocate / use low-level variables. And there would have to be some kind of guarantee that the two states stay separated.

oleganza:
Probably no need for separate Builder type: have `prove` method consume self.

hdevalence:
The `prove` method does consume `self` already, but that doesn't deal with this issue, which is that the high-level variables (constraint system inputs) need to all be set before the constraints are added (which has to happen before proving).

oleganza:
Oops, you are right. Then we can have `Prover` as a starting type which is then converted to `ProvingCS` by the `prove` method that takes a closure with CS as an argument, consumes the prover and returns the proof (the Prover being the “builder”). Closure would let us avoid having two steps “convert to CS” and “make proof”.

oleganza:
Fixed in #221.

PR #221, opened 2018-11-27
Allow committing high-level variables one-by-one

oleganza:
Addresses #218.

The API entry points are now `Prover` and `Verifier` types that let you commit high-level variables individually, so that one can organize resulting variables in the way matching the outer protocol (e.g. in tuples). 

The `ProverCS` and `VerifierCS` are provided to the user when `Prover` and `Verifier` are consumed, which ensures that high-level variables cannot be committed during CS construction.

Building a CS happens inside a closure provided to `prove`/`verify` methods, so the user effectively "passes a CS" as an argument to the `prove` or `verify`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ criterion = "0.2"
 bincode = "1"
 
 [features]
+default = ["yoloproofs"]
 avx2_backend = ["curve25519-dalek/avx2_backend"]
 yoloproofs = []
 
--- a/benches/r1cs.rs
+++ b/benches/r1cs.rs
@@ -1,5 +1,5 @@
 extern crate bulletproofs;
-use bulletproofs::r1cs::{ConstraintSystem, ProverCS, R1CSError, R1CSProof, Variable, VerifierCS};
+use bulletproofs::r1cs::{ConstraintSystem, Prover, R1CSError, R1CSProof, Variable, Verifier};
 use bulletproofs::{BulletproofGens, PedersenGens};
 
 #[macro_use]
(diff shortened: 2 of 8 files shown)


oleganza:
@cathieyun oh, cool! then we'd have separate pair of eyes trying to actually use this API.

oleganza:
Re: "why closures?"

Current API looks roughly like this:

```rust
let prover = Prover::new(...);
let variable = prover.commit(v, v_b);
let proof = prover.prove(|cs| gadget(cs, variable));

let verifier = Verifier::new(...);
let variable = verifier.commit(V);
let result = verifier.verify(|cs| gadget(cs, variable));
```

Without using closures we need another method that converts Prover/Verifier into CS:

```rust
let prover = Prover::new(...);
let variable = prover.commit(v, v_b);
let cs = prover.into_cs();
gadget(cs, variable);
let proof = cs.prove();

let verifier = Verifier::new(...);
let variable = verifier.commit(V);
let cs = verifier.into_cs();
gadget(cs, variable);
let result = cs.verify();
```

Which one do you like better? And what could be a better name than `into_cs`?

cathieyun:
(discussed offline): I slightly prefer the latter, no-closure approach, because it's a little clearer what is happening at each step. Maybe `into_cs` could be something like `finish_receiving_commitments` or `start_adding_constraints` or something like that which denotes the transition from the receiving commitments stage to the adding constraints stage.

oleganza:
After talking to @cathieyun, will change the API to a more straightforward sequence of calls. Will rename `into_cs()` into `build_constraint_system()`.

cathieyun:
Working on the spacesuit update here:
https://github.com/interstellar/spacesuit/pull/19

oleganza:
One interesting consequence of this, is that we cannot add `m` into the transcript before all high-level variables are added there (since they are not added all at once), so it becomes more like a "suffix" (i have added a comment on why it's still safe). 

Alternatively, we can remember all high-level variables in an array and dump them altogether into the transcript exactly as we did previously. But this would require having that extra array.

oleganza:
Henry:
> what about `finalize_inputs` instead of `build_constraint_system`?
> ...
> the idea was just that the name for the state transition should describe what the state transition is

Renamed in 2b28cd9.

Issue #222, opened 2018-11-29
Two-phase CS construction with intermediate commitments (cont)


oleganza:
(This is a reset of the discussion in #186 and proof-of-concept #196, both of which got derailed by random implementation ideas.)

### Implementation

Docs: https://github.com/dalek-cryptography/bulletproofs/pull/224
Code: https://github.com/dalek-cryptography/bulletproofs/pull/232

### Problem

Our current R1CS interface allows constructing constraints using challenges bound to the external variables (**V**). This, however, may be risky for complex circuits. For example, a _shuffle gadget_ is only binding if inputs on both sides are committed. However, if one side remains uncommitted (e.g. when there is a circuit in between two shuffles and only "outer" sides are ultimately committed), then the low-level variables on that side can be forged based on the knowledge of the challenge.

### Goal

Make gadgets _locally sound_ by giving an opportunity to commit to input low-level variables, regardless of whether they are indirectly committed via high-level variables.

### Proposal

We split the circuit construction in two phases.

**Phase 1**

Gadgets can allocate low-level variables, add constraints, but cannot use challenges. This phase produces first pair of vector commitments `A_I1, A_O1, S1` to these variables.

This allows allocating all the gadgets and all variables between them, assigning required values to these variables. But not all gadgets can implement their internal constraints yet because they need challenge values.

The phase ends with three points `A_I1, A_O1, S1` computed and sent to the transcript, committing to all the low-level variables allocated so far.

**Phase 2**

Gadgets can allocate more low-level variables, add more constraints _and can generate challenges_. This phase produces second set of vector commitments `A_I2, A_O2, S2` to these additional low-level variables.

This allows implementing gadgets, such as _shuffle_ or _merge_ using challenges, making them **locally sound** because all input variables for any given gadget are already committed via `A_I1,A_O1,S1` in the first phase.

### How this works?

The goal is to make the prover commit to the first-phase variables _before_ the challenges are generated and used to compute assignments for the second-phase variables and weights.

This means that we need _all statements_ about the first-phase variables to be true _before_ the second-phase variables are computed.

Lets look at the statements between `l,r` and `a_R,L,O`. Consider that this is a n-item vector of _pairs_ of statements between items in `l[i],r[i]` and each multiplicate gate's wires `a[i]_R,L,O`.

<img width="444" alt="image" src="https://user-images.githubusercontent.com/698/49190002-60501280-f325-11e8-9090-9a013070603f.png">

[...]

PR #232, opened 2018-12-14
Implement API for two-phase constraint system proofs

oleganza:
This provides the implementation for two-phase protocol redesign as described in #222.

Challenges are available in a section of code within a closure provided to `cs.specify_randomized_constraints` call:

```rust
cs.specify_randomized_constraints(move |cs| {
    let z = cs.challenge_scalar(b"shuffle challenge");
    ...
})
```

Prover and Verifier now allow interspersing high-level variable commitments with the constraints in the first phase (which allows gadgets to introduce their own high-level variables during CS construction).

Since the callbacks must return a known-to-the-framework error type (`R1CSError`) and some gadgets can be composed within the second phase (cf. tuple shuffle + scalar shuffle), this PR also introduces a catch-all error variant `R1CSError::GadgetError` which allows gadgets to signal inconsistency in their parameters.


src/r1cs/constraint_system.rs
```diff
@@ -53,19 +53,17 @@ pub trait ConstraintSystem {
     /// ```
     fn constrain(&mut self, lc: LinearCombination);
 
-    /// Obtain a challenge scalar bound to the assignments of all of
-    /// the externally committed wires.
+    /// Randomize the constraints using a randomly sampled challenge scalar
+    /// bound to the assignments of the non-randomized variables.
```
hdevalence: Saying "randomize the constraints" sounds like there are some existing constraints, and then, these existing constraints are randomized, when we really mean that this is a hook for adding constraints which depend on a random value.

I'm not sure exactly what would be better, but maybe this could be explained in terms of the role of the callback?  Right now the role of the `Fn(&mut Self, Scalar)` is a little unclear. We should probably say that the callback is supposed to use the random input scalar to construct further constraints.

Also, is there a more precise name for what we're doing than `randomized_constraints`? It feels like it's still not quite right, because it's not that we're adding randomized constraints in *this* call, which is a hook to call the closure to add them later.

oleganza: `define_randomized_constraints`?

src/r1cs/verifier.rs
```diff
@@ -94,8 +86,24 @@ impl<'a, 'b> ConstraintSystem for VerifierCS<'a, 'b> {
...
+        match self.deferred_constraints {
+            Some(ref mut list) => {
+                list.push((label, Box::new(callback)));
+                return Ok(());
+            }
+            None => {}
+        }
+
+        let challenge = self.transcript.challenge_scalar(label);
+        callback(self, challenge)
```
hdevalence: Shouldn't this be part of the `None` arm of the match?

oleganza: Yup. Will change.

oleganza: Fixed in a22c3d0

src/r1cs/verifier.rs
```diff
@@ -94,8 +86,24 @@ impl<'a, 'b> ConstraintSystem for VerifierCS<'a, 'b> {
...
+        &mut self,
+        label: &'static [u8],
+        callback: F,
+    ) -> Result<(), R1CSError>
+    where
+        F: 'static + Fn(&mut Self, Scalar) -> Result<(), R1CSError>,
+    {
+        match self.deferred_constraints {
+            Some(ref mut list) => {
+                list.push((label, Box::new(callback)));
```
hdevalence: I'm not sure I understand what's happening here... if
```
callback: F,
F: 'static + Fn(&mut Self, Scalar) -> Result<(), R1CSError>,
```
then why do we need to box `callback`? Isn't it already `'static` and will therefore outlive the `Verifier`?  Alternatively, if we are going to box it, does it need to be `'static`?

src/r1cs/constraint_system.rs
oleganza: Renamed to `specify_randomized_constraints` and added more API documentation in cced8b7

src/r1cs/verifier.rs
oleganza: Hmm. Good question. I think this is an artifact of me-not-knowing-Rust-well tinkering in #196 and ending up with something that finally compiles.

I guess i'll try to see if we can keep `'static` and avoid boxing?

src/r1cs/verifier.rs
```diff
@@ -47,9 +36,20 @@ pub struct VerifierCS<'a, 'b> {
     /// variable assignments.
     num_vars: usize,
     V: Vec<CompressedRistretto>,
+
+    /// This list holds closures that will be called in the second phase of the protocol,
+    /// when non-randomized variables are committed.
+    /// After that, the option will flip to None and additional calls to `randomize_constraints`
+    /// will invoke closures immediately.
```
hdevalence: This behavior (i.e., that calls to `randomize_constraints` invoke closures immediately in phase 2) seems important to mention in the user-facing docs, because it's what allows multiple challenges to be used for a set of constraints. I don't know how much we want to emphasize it, since doing something like
```
cs.specify_randomized_constraints(|cs, z| {
    cs.specify_randomized_constraints(|cs, w| {
        /* z, w = z, w */
    }
}
```
seems worse than 
```
cs.specify_randomized_constraints(|cs, z| {
    /* z, w = z, z*z */
}
```

oleganza: `z, w = z, z*z` is not safe, because if you need to generate powers of `z` and `w` down the road, those powers can overlap (see the use of `y` and `z` in the core bulletproofs protocol).

src/r1cs/verifier.rs
oleganza: Seems like we need to do `Box<'static+Fn...>`. Here's why:

1. Closures are all of different (anonymous) type as they are created by distinct gadgets, but we need to store them in a `Vec`, so they have to be wrapped in a single type. Hence, boxing. In other words, `Fn` is a trait, not a type.
2. Closure is owned by box, which erases lexical lifetime and therefore requires `'static`. This, in turn, implies that variables can be only captured by move into closure, not by reference (unless the references are 'static themselves).

src/r1cs/verifier.rs
oleganza: Alternative, is to provide a different (sub)trait for CS which has method `challenge_scalar(label)->Scalar` which the user can call themselves.

oleganza: So it'll look something like this:

```
cs.specify_randomized_constraints(|cs| {
    let x = cs.challenge_scalar("x");
    let y = cs.challenge_scalar("y");
    ...
}
```

oleganza: I wanted to try an associated type `type RandomizedCS` set to Self in Prover/Verifier. I can check if it needs to be in scope.

oleganza: it might not only be cleaner in code, but also conceptually clean: the closure is actually specifying a _phase_, in which one can sample one or more challenges, so doing both of these things as separate calls matches this closer.

hdevalence: Oh, a complication with doing this is that right now we keep the phase1/phase2 state internal to the CS impl (specifically the last function where it's consumed).  But if we want to have the `cs` parameter to the constraint callback implement a `Phase2CS` trait, it needs to be a different type so that the trait can be available only to that parameter and not on the original constraint system.

oleganza: Did some code changes: turns out, the user does not need to bring in scope RandomizedConstraintSystem to use challenges. And to your point, if the same type implements both traits, then user can define gadget over `<CS:RandomizedConstraintSystem>` which would be compatible with sending Prover/Verifier there (although it can't be used from within a correctly parametrized gadget in first-phase)?

oleganza: We can easily add another type by wrapping `Prover` in `RandomizingProver` and unwrapping after we are done with second phase.

oleganza: This would also make `mem::replace` and `Option<Vec>` go away.

src/r1cs/verifier.rs
hdevalence: Cool, requiring a trait object to stick them in a `Vec` makes sense.

src/r1cs/verifier.rs
hdevalence: I think that the same type should *not* implement both traits, because it breaks the guarantees we achieve with this API (e.g., being able to allocate HL variables while forming constraints with no problems)...

oleganza: yeah, that's what i meant above - having two separate types for pre-randomization and randomization phases.

hdevalence: Here's a sketch of a design that I think would work: we add a trait
```
pub trait Phase2CS: ConstraintSystem {
    /// Request a challenge scalar with the given domain separation label                                                                                                                                          
    fn challenge_scalar(&mut self, label: &'static [u8]) -> Scalar;
}
```
and have the callback take `cs: Phase2CS`.  Then we create (non-public) types `Phase2Prover` and `Phase2Verifier`, with `From` impls from `Prover` and `Verifier`.  The proving / verification logic moves into the `Phase2` types. When a user calls `Prover::prove()`, the method consumes the `self: Prover`, moves the data into a `Phase2Prover`, which executes the callbacks, passing itself (as a `Phase2CS` impl) to the callbacks, producing a proof (or verification) and then immediately returning the `Result`.

oleganza: yeah, i'm half-way doing that, but these types should be made public (although with private field, so only our crate can instantiate them).

oleganza: WIP here: https://github.com/dalek-cryptography/bulletproofs/compare/oleg/two-phase-impl...oleg/two-phase-impl-two-traits?expand=1

hdevalence: Why should they be public?

oleganza: Looks like, they need to be public so that compiler can monomorphize the closures. They are anonymous types, but they need to be fully known nonetheless.

The good news is: the user does not have to even bring those in scope, and they cannot do anything with them outside the context of that `specify_randomized_constraints` call.

src/r1cs/constraint_system.rs
```diff
@@ -16,6 +16,10 @@ use curve25519_dalek::scalar::Scalar;
 /// using the `ConstraintSystem` trait, so that the prover and
 /// verifier share the logic for specifying constraints.
 pub trait ConstraintSystem {
+    /// Represents a concrete type for the CS in a randomization phase.
+    /// Both prover and verifer implement both traits, so this type equals Self.
```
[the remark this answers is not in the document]
oleganza: Oops, it’s a stale comment from the past iteration. Will fix.

hdevalence:
Hmm, how does breadth-first evaluation work with nested callbacks? At least to me the previous method seemed simpler to understand (even if it meant that adding the callback could not be infallible)..

cathieyun:
Note: with the new error type, it is not necessary to make the callbacks infallible, since the nested gadgets can now handle errors properly.

oleganza:
> how does breadth-first evaluation work with nested callbacks? At least to me the previous method seemed simpler to understand (even if it meant that adding the callback could not be infallible).

The previous method had two behaviors: asynchronous if you are in the first phase, and immediate if you are in the second. Consequently, internal implementation would be calling the callbacks from two different places.

The new implementation says "the callback is always deferred", and the callbacks are all called from the single place (although with a loop to support multiple passes).

oleganza:
>Note: with the new error type, it is not necessary to make the callbacks infallible, since the nested gadgets can now handle errors properly.

Yeah, but still it seems cleaner to me to have one behavior instead of two for the same method, and that behavior never fails.

hdevalence:
Looking at the diff https://github.com/dalek-cryptography/bulletproofs/pull/232/files/1e9c29b8e520389d112145385d4f1101ca02d46c..f188e73ce245a4d7e0a382653f4bdd0a9303f889 , it seems like the change to the breadth-first evaluation makes things more complicated, and it's less clear how the callbacks will be executed.  It seems simpler to use the original design, especially since the original reason to consider changing it is now gone.

hdevalence:
Where are the "phase 1" commitments formed and committed to the transcript? Am I missing something?

hdevalence:
(oops, hit the wrong button)

oleganza:
>Where are the "phase 1" commitments formed and committed to the transcript? Am I missing something?

Oops. Pushed the update.

oleganza:
>it seems like the change to the breadth-first evaluation makes things more complicated, and it's less clear how the callbacks will be executed. It seems simpler to use the original design, especially since the original reason to consider changing it is now gone.

Ok. Reverted.

oleganza:
@cathieyun i've fixed the bug that you triggered in spacesuit: 
https://github.com/dalek-cryptography/bulletproofs/pull/232/commits/7dd7081b9381de5731e3ced443031b2a02497586

So the adjustment factors for generators G were not swapped as needed, and our shuffle tests were not checking for that because all low-level variables were in the second phase, making the adjustment factors array uniform: `[u,...,u]`. While in the Cloak's padded shuffle, some low-level variables were in the first phase (specifically, the padding variables), making the factors array non-uniform: `[0,...,0,u,...,u]`, so when its halves are swapped, the result is not the same.

Will need to add a regression test for this.

cathieyun:
The changes look good! It would be good to add the cloak padded shuffle (or some example that tests that the adjustment factors are correct) to this PR - I can work on that

oleganza:
💖

oleganza:
Follow-up discussion on making this thing optional: https://github.com/dalek-cryptography/bulletproofs/issues/242

PR #239, opened 2019-01-02
Typo in notes-r1cs.md

roconnor-blockstream:


docs/notes-r1cs.md
```diff
@@ -280,7 +280,7 @@ Rearrange \\(\langle \mathbf{a}\_L \circ \mathbf{a}\_R, \mathbf{y}^n \rangle\\)
 \\[
 w\_c + \langle \mathbf{w}\_V, \mathbf{v} \rangle
 =
-\langle \mathbf{a}\_L \circ \mathbf{a}\_R, \mathbf{y}^n \rangle +
+\langle \mathbf{a}\_L, \mathbf{y}^n \circ \mathbf{a}\_R \rangle +
```
hdevalence: Is this a typo? I think that the before and after expressions are equivalent, right?

cathieyun: Agreed - the line is not _incorrect_, but it fails to actually reflect the change (rearrange) it's supposed to.

PR #241, opened 2019-01-15
Refactor k-shuffle benchmark

### LINKED ISSUE #240: R1CS: benchmark with `bench_function_over_inputs`
Hi! Cool work on R1CS! Experiments only, I promise :-)

I was looking at the benchmarks, and noticed it was run for multiple values without using criterion's `bench_function_over_inputs`. Is there a particular reason for this? If not, I can send a PR that refactors that.

rubdos:
Would fix #240. If this is not wanted, no big deal.

### DIFF
--- a/benches/r1cs.rs
+++ b/benches/r1cs.rs
@@ -144,7 +144,7 @@ impl KShuffleGadget {
             .map(|v| prover.commit(*v, Scalar::random(&mut blinding_rng)))
             .unzip();
 
-        let cs = prover.finalize_inputs();
+        let mut cs = prover.finalize_inputs();
 
         Self::fill_cs(&mut cs, &input_vars, &output_vars);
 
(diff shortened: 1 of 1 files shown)


cathieyun:
This looks good to me, it's just missing a `cargo fmt`.
Just to check, this addressed the first part of #240 (regarding `bench_function_over_inputs` instead of `bench_function`) but not the second, right?

rubdos:
>  Just to check, this addressed the first part of #240 (regarding `bench_function_over_inputs` instead of `bench_function`) but not the second, right?

I've fixed both in separate commits. If you want, you can cherry pick what you need.

> This looks good to me, it's just missing a `cargo fmt`.

`cargo fmt` did change it quite a bit :-/

cathieyun:
I'll merge this in, but will end up changing quite a bit when I rebase this PR off of it:
https://github.com/dalek-cryptography/bulletproofs/pull/232
(That PR already fixes the mutable constraint system problem as a byproduct of the 2-phase protocol.)

Issue #242, opened 2019-01-16
Optional API for challenge-based R1CS proofs


oleganza:
(This is a summary of an issue raised by @hdevalence in an offline discussion.)

The challenge-based constraints in R1CS API (#222) are pretty novel and not have a formal security proof yet. It makes sense to make the use of those _optional_ with respect to the rest of the R1CS API.

### Proposal

First, it should be possible for the verifier to ensure that randomization API is not available throughout the hierarchy of gadgets unless it was explicitly turned on. This makes challenge-based circuits an "opt-in" feature.

The proofs for non-randomized constraints should be compatible with the proofs for randomizable constraints: that is, they could be verified by a less conservative verifier, as well as a conservative one. For this we'd simply need to pad the transcript with dummy zero commitments `(A_I2, A_O2, S2)`. This padding should obviously not affect the correctness of the original R1CS protocol.

### Questions

1. Do we have different proof types for with randomization and without randomization; or do we use the uniform type with a flag distinguishing the two?
2. Do we try to save 96 bytes of all-zero second-phase commitments `(A_I2, A_O2, S2)` for the proofs of non-randomized CS, or just have a uniform proof format?
3. Do we turn on/off the randomization API dynamically or statically? E.g. is a subtrait of a `RandomizableConstraintSystem: ConstraintSystem`? Gadgets that do not use randomization could use a more conservative supertrait, but gadgets that need it would use a subtrait.

WDYT? @hdevalence @cathieyun

oleganza:
The minimal support for compact proofs would be:

1. One type `R1CSProof` as we have right now.
2. When serializing, check if the 3 second-phase commitments are `.is_identity()`, then avoid encoding them altogether.
3. When deserializing, check if the byteslice is odd - in this case replace missing 3 points with 3 identity points.
4. When making a proof, make sure these commitments are left as identity points (with zero blinding factors) in case there are no second-phase variable allocations.

hdevalence:
> The prover/verifier knows whether they are going to use randomization (that's baked into the circuit they use), but they know it _late_: when the gadgets actually trigger randomization. @hdevalence's suggestion is to look at the size of the proof bytestring and if it's even, it's a 2-phase proof, if it's odd - it's a 1-phase proof (16+2k elements VS 13+2k elements). This way we can avoid having a special flag.

[...]

PR #243, opened 2019-01-17
Implements serialization of R1CS

### LINKED ISSUE #214: Implement serialization for `R1CSProof`s
The `R1CSProof` should have a serialization format; the `RangeProof` code could be a good example.

rubdos:
Fixes #214 

~~I'm writing a roundtrip test, will update soon with a second commit.~~ Also ran cargo fmt on the respective commits.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -97,6 +97,9 @@ pub enum R1CSError {
     /// Occurs when there are insufficient generators for the proof.
     #[fail(display = "Invalid generators size, too few generators for proof")]
     InvalidGeneratorsLength,
+    /// This error occurs when the proof encoding is malformed.
+    #[fail(display = "Proof data could not be parsed.")]
+    FormatError,
     /// Occurs when verification of an
     /// [`R1CSProof`](::r1cs::R1CSProof) fails.
     #[fail(display = "R1CSProof did not verify correctly.")]
--- a/src/r1cs/proof.rs
+++ b/src/r1cs/proof.rs
@@ -1,9 +1,15 @@
+#![allow(non_snake_case)]
 //! Definition of the proof struct.
 
 use curve25519_dalek::ristretto::CompressedRistretto;
 use curve25519_dalek::scalar::Scalar;
 
+use errors::R1CSError;
 use inner_product_proof::InnerProductProof;
+use util;
+
+use serde::de::Visitor;
+use serde::{self, Deserialize, Deserializer, Serialize, Serializer};
 
 /// A proof of some statement specified by a
 /// [`ConstraintSystem`](::r1cs::ConstraintSystem).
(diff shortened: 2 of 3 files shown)


rubdos:
Maybe useful to add to the docs of `to_bytes` is the output size of the serialization. I can count the Ristretto points and scalars, but I'm not quite sure how to compute the size of the r1cs ipp in function of the circuit and variables.

oleganza:
Maybe these notes might help? https://doc-internal.dalek.rs/develop/bulletproofs/r1cs/notes/index.html

image

Note: the doc is outdated - it's actually 16+2k because of extra 3 points due to 2-phase design.

rubdos:
Ha, I had run through these notes, but have never seen these paragraphs -- weird. I will add the comment tomorrow morning.

rubdos:
With the output size fix:

screen shot 2019-01-18 at 10 17 52

src/r1cs/proof.rs
```diff
@@ -56,3 +62,135 @@ pub struct R1CSProof {
...
+
+impl R1CSProof {
+    /// Serializes the proof into a byte array of \\(16 + 2k\\) 32-byte elements,
+    /// where \\(k=\lceil \log_2(n) \rceil\\) and \\(n\\) is the number of multiplication gates.
+    ///
+    /// # Layout
+    ///
+    /// The layout of the r1cs proof encoding is:
+    ///
+    /// * eleven compressed Ristretto points \\(A_I1,A_O1,S1,A_I2,A_O2,S2,T_1,...,T_6\\),
```
oleganza: Fixing subscripts:
```suggestion
    /// * eleven compressed Ristretto points \\(A_{I1},A_{O1},S_1,A_{I2},A_{O2},S_2,T_1,...,T_6\\),
```

src/errors.rs
```diff
@@ -97,6 +97,9 @@ pub enum R1CSError {
     /// Occurs when there are insufficient generators for the proof.
     #[fail(display = "Invalid generators size, too few generators for proof")]
     InvalidGeneratorsLength,
+    /// This error occurs when the proof encoding is malformed.
+    #[fail(display = "Proof data could not be parsed.")]
+    FormatError,
```
oleganza: Since we do not decompress points or check scalars until much later in the verification code, the FormatError is simply an error in size of a proof. IMHO it does not provide useful diagnostic information, so we can keep using the plain "ProofError" like elsewhere in the code. Also, users cannot really recover from this error, so if the proof is invalid (for whatever reason), the whole user's protocol should be aborted the same way (cf. recent [PGP efail attack](https://efail.de)).

rubdos: I felt like `ProofError` was `RangeProof`-specific, but it indeed felt awkward to add this. Will add a separate commit, and if it's indeed better I'll `fixup` the first commit.

oleganza: > you now would need to convert from both ProofError and R1CSError to your own Error format if you're writing a program.

This seems to be the case no matter what error types we use in-library, because if the user has a different error type, they'd have to convert or wrap this one anyway.

oleganza: (deleted previously posted incorrect comment)

I don't think you'd often have a code that uses both RP and R1CS APIs. Because if you already use R1CS, then it's cheaper to just use rangeproof-as-a-gadget within R1CS, w/o using a dedicated rangeproof API at all.

rubdos: Indeed my point. `ProofError` has `RangeProof`-specific values, so if the deserialization of `R1CSProof` would return a `ProofError`, a R1CS user would need to handle the `RangeProof`-error values, which doesn't make sense.

Do you still want me to return `ProofError`?

oleganza: Oh, I apologize for confusion I caused. Originally I meant R1CSError when i said ProofError.

oleganza:
Follow-up discussion: 
https://github.com/dalek-cryptography/bulletproofs/issues/242#issuecomment-457019429

Probably it's a good idea for a separate PR after this one.

rubdos:
> Follow-up discussion:
> [#242 (comment)](https://github.com/dalek-cryptography/bulletproofs/issues/242#issuecomment-457019429)
> 
> Probably it's a good idea for a separate PR after this one.

Since what you suggest in that comment would be backwards compatible, I guess we're round here? Except if you still would want some change to the error system that is.

Thanks for the support and checks!

src/r1cs/proof.rs
```diff
@@ -56,3 +62,135 @@ pub struct R1CSProof {
...
+    /// Returns the size in bytes required to serialize the `R1CSProof`.
+    pub fn serialized_size(&self) -> usize {
+        // 14 elements + the ipp
+        14 * 32 + self.ipp_proof.serialized_size()
+    }
+
+    /// Deserializes the proof from a byte slice.
+    ///
+    /// Returns an error if the byte slice cannot be parsed into a `R1CSProof`.
+    pub fn from_bytes(mut slice: &[u8]) -> Result<R1CSProof, R1CSError> {
```
oleganza: This has the same trailing-bytes issue as RangeProof, but we can fix both in the follow-up PR. Issue filed here: https://github.com/dalek-cryptography/bulletproofs/issues/255

Issue #198, opened 2018-11-06
Non-power-of-two inner product proof


oleganza:
**Problem**

Inner product proof requires inputs with power-of-two items. This could hurt performance of proving and verification up to 2x on input sizes that are just-a-bit-larger that a power of two.

For example, a confidential assets scheme has 8+64 multipliers per transfer, where 8 are used for mixes and shuffles, and 64 are for the range proof. This means, the closest power of two is exceeded by merely 11%, making the padding increase the cost almost two-fold.

**Suggestion**

We can modify the IPP in a compatible manner (that is, the power-of-two IPP remains the same) by padding the inputs with _zero scalars_ `0` and _identity points_ `O`. This padding does not change the original relation and, due to the use of challenges when folding the lo/hi halves of the vectors, does not interfere with other scalars or generators.

```
P = <a, G> + <b, H> + <a,b> Q = 
  = <a, G> + <0, O> + <b, H>  + <0, O> + <a,b> Q + <0,0> Q
```

This way, the verifier does not have to include operations on the identity points in the megacheck, and therefore removes the overhead.

To make it similarly more efficient for the _prover_, we can rearrange the padding in the following manner:

1. Let `n` be size of inputs and `k = ceil(log_2(n))` the next power of two.
2. Pad the inputs to the _even_ size `n' = 2*ceil(n/2)`. In other words, add `0` and `O` to the scalars and generators accordingly, if the input size is odd. Otherwise, do nothing.
3. Split each input vector in half.
4. Pad each resulting vector to an even size `n'' = 2*ceil(n'/2)` again (if needed).
5. We can proceed until we get to the size one. This will effectively pad the input with interspersed zeroes.

```
18-item input to be padded to 32-item input:
[a b c d e f g h i j k l m n o p q r]
```

We need 14 zeroes which we will arrange as `2*(4+2+1)`:

```
a b c d e f g h i j k l m n o p q r

already padded:

a b c d e f g h i j k l m n o p q r

split in half and fold (9-item vector of [a+j, b+k, ...]):

a b c d e f g h i
j k l m n o p q r

pad to even:

a b c d e f g h i 0
j k l m n o p q r 0

split in half and fold (5-item vector of [a+j+f+o, b+k+g+p, ...]):

a b c d e
j k l m n
f g h i 0
o p q r 0

pad to even:

a b c d e 0
j k l m n 0
f g h i 0 0
o p q r 0 0

split in half and fold:

a b c
j k l
f g h
[...]

PR #252, opened 2019-02-08
Add benchmarks for BulletproofGens and PedersenGens

rubdos:
I noticed BulletproofGens is quite slow with larger generator sizes. This commit adds a benchmark for both generators. Figured it may come in handy some day.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,6 +47,10 @@ name = "range_proof"
 harness = false
 
 [[bench]]
+name = "generators"
+harness = false
+
+[[bench]]
 name = "r1cs"
 harness = false
 required-features = ["yoloproofs"]
--- a/benches/generators.rs
+++ b/benches/generators.rs
@@ -1,0 +1,26 @@
+extern crate bulletproofs;
+use bulletproofs::{BulletproofGens, PedersenGens};
+
+#[macro_use]
+extern crate criterion;
+use criterion::Criterion;
+
+fn pc_gens(c: &mut Criterion) {
+    c.bench_function("PedersenGens::new", |b| b.iter(|| PedersenGens::default()));
+}
+
+fn bp_gens(c: &mut Criterion) {
+    c.bench_function_over_inputs(
+        "BulletproofGens::new",
+        |b, size| b.iter(|| BulletproofGens::new(*size, 1)),
+        (0..10).map(|i| 2 << i),
+    );
+}
+
+criterion_group! {
+    bp,
+    bp_gens,
+    pc_gens,
+}
+
+criterion_main!(bp);


oleganza:
Thank you!

Yeah, making generators is slow, that's why APIs allow you to pass a precomputed instance around, so you can reuse it as long as the process lives.

Do you mind running `cargo fmt` to make Travis CI happy?

rubdos:
> Yeah, making generators is slow, that's why APIs allow you to pass a precomputed instance around, so you can reuse it as long as the process lives.

Makes me wonder; would it be useful to have some standard-sized precomputed at compile time? That way, we don't need to initialize them on potentially slower devices.

rubdos:
Maybe I should add some benches for the internal generators too?

oleganza:
>Makes me wonder; would it be useful to have some standard-sized precomputed at compile time? That way, we don't need to initialize them on potentially slower devices.

Is it slow enough to make booting up a process uncomfortably slow? If you use a small device as a network node that validates a lot of proofs, then it can create an instance of generators with a necessary capacity upfront and then reuse it. Does it matter if it took a second or two to generate it?

oleganza:
>Maybe I should add some benches for the internal generators too?
WDYM?

rubdos:
> Is it slow enough to make booting up a process uncomfortably slow? If you use a small device as a network node that validates a lot of proofs, then it can create an instance of generators with a necessary capacity upfront and then reuse it. Does it matter if it took a second or two to generate it?

Nah, I suppose not. If it does hurt me some day, I'll come back to it.

>> Maybe I should add some benches for the internal generators too?
> WDYM?

Meant to add a benchmark for `AggregatedGensIter`, but that doesn't actually compute anything significant.

For me this is finished for now :-)

oleganza:
On my 3.1 GHz Intel Core i7, I get **256** generators created in **5.1 ms**. This means, something like a SHA256 circuit (~25K multiplication gates => 25K generator pairs) would take **0.5 sec**. Even if it's 2-3x slower, it's still pretty fast for a one-time boot cost even for such a big circuit.

At the same time, the cost of validating each proof with such a circuit size would be higher than making the necessary amount of generators, so the marginal cost of creating them upfront becomes negligible after just a handful of proofs.

oleganza:
I've changed the base branch to `develop`, this is where all the new patches are going to. `main` contains stable releases only.

rubdos:
> I've changed the base branch to `develop`, this is where all the new patches are going to. `main` contains stable releases only.

Woop, sorry! My bad; I knew that :-)

rubdos:
> On my 3.1 GHz Intel Core i7, I get **256** generators created in **5.1 ms**. This means, something like a SHA256 circuit (~25K multiplication gates => 25K generator pairs) would take **0.5 sec**. Even if it's 2-3x slower, it's still pretty fast for a one-time boot cost even for such a big circuit.

docs say:
> `gens_capacity` is the number of generators to precompute for each party. For rangeproofs, it is sufficient to pass 64, the maximum bitsize of the rangeproofs. For circuit proofs, the capacity must be greater than the number of multipliers, rounded up to the next power of two.

So for 25k, that gives a `gens_capacity` of 32768. Criterion wants to take >1h to benchmark this... Anyhow: my threadripper measures 739.08 ms ± 6ms.

oleganza:
Oh, i forgot about rounding up. But on the other hand, we can fix it: https://github.com/dalek-cryptography/bulletproofs/issues/198

PR #253, opened 2019-02-13
Fix typo in r1cs notes

rubdos:


rubdos:
Second commit (c6fe44f) does not fix a typo, but I figured you folks might want that too. It commutates `B.v` to `v.B` in the notes at four points, such that the EC base point is at the right hand of the multiplication, as is in the rest of the notes.

PR #256, opened 2019-03-08
r1cs: single-variable allocation API + simpler API for assignments

### LINKED ISSUE #206: API for allocating one low-level variable at a time
Currently the API `assign_uncommitted` exposes a whole multiplier, returning two variables: left and right wire. If the user needs only one, they can discard the other. This is slightly suboptimal and looks like a leaky abstraction: in a loop over K gadgets that are connected with K-1 "wires", the user is faced with a dilemma — try to unroll the loop in pairs to avoid throwing out half of variables, or ignore them?

The suggestion is:

1. The API always returns one variable at a time.
2. If there is no half-assigned multiplier: CS allocates a new one, assigns and returns the left wire, assigns 0 to the unused right wire and keeps track of the multiplier till the next call, or till the CS is committed.
3. When the next call comes in, the remembered multiplier's right wire is assigned, output wire is recalculated, right wire is returned to the user, and CS forgets about that multiplier.
4. When CS is committed, if there is a half-assigned multiplier hanging on, CS "forgets" it and keeps its zero-assignments to the right and output wire.

This way, each odd call creates a new multiplier and each even call completes the assignment of the already existing multiplier. If CS has to allocate the normal multiplier in the middle, it does not conflict with the half-assigned multiplier because it would have its own index in the array from the beginning.

This change should be pretty small in its complexity. I don't think it is a meaningful performance win, but the API should be more clean this way.

PS. Per https://github.com/dalek-cryptography/bulletproofs/issues/186, first-phase variables are no longer "uncommitted" and can very well be committed. So we should find a more appropriate name, e.g. "assign_internal_variable" or something like that.

PR: #256

oleganza:
This adds efficient and clean API for allocating one variable at a time.

Here is how it simplifies the code in Spacesuit and ZkVM: https://github.com/interstellar/slingshot/pull/211

Plus, the assignments are provided simply as `Option<Scalar>` and `Option<(Scalar,Scalar)>` to make code look nicer. The overhead in verifier of checking for `None` and not doing `Option::map` or something in the gadget code should be negligible.

Closes #206

### DIFF
--- a/src/r1cs/constraint_system.rs
+++ b/src/r1cs/constraint_system.rs
@@ -46,7 +46,7 @@ pub trait ConstraintSystem {
     /// ```
     ///
     /// Returns `(left, right, out)` for use in further constraints.
-    fn allocate<F>(&mut self, assign_fn: F) -> Result<(Variable, Variable, Variable), R1CSError>
+    fn allocate_multiplier<F>(&mut self, assign_fn: F) -> Result<(Variable, Variable, Variable), R1CSError>
     where
         F: FnOnce() -> Result<(Scalar, Scalar, Scalar), R1CSError>;
 
--- a/src/r1cs/constraint_system.rs
+++ b/src/r1cs/constraint_system.rs
@@ -41,7 +41,7 @@ pub trait ConstraintSystem {
 
     /// Allocate a single variable.
     ///
-    /// This either allocates a new multiplier and returns it’s `left` variable,
+    /// This either allocates a new multiplier and returns its `left` variable,
     /// or returns a `right` variable of a multiplier previously allocated by this method.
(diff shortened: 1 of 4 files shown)


oleganza:
This API is used here: https://github.com/interstellar/slingshot/pull/211

src/r1cs/constraint_system.rs
```diff
@@ -39,16 +39,29 @@ pub trait ConstraintSystem {
         right: LinearCombination,
     ) -> (Variable, Variable, Variable);
 
+    /// Allocate a single variable.
+    ///
+    /// This either allocates a new multiplier and returns it’s `left` variable,
```
oleganza:
```suggestion
    /// This either allocates a new multiplier and returns its `left` variable,
```

PR #257, opened 2019-03-11
r1cs: provide BP generators later in the proving/verifying workflow

oleganza:
This change allows users to provide Bulletproofs generators closer to the point where they are needed. This allows a prover, for instance, to know exactly how many multipliers were allocated to prepare a necessary number of generators, and also specify this number in the envelope that holds the proof. Verifier then also has an option to generate a necessary amount of generators based on actual usage, or preallocate as many as specified in the envelope (if not determined by some other factors).

For the verifier, the same thing is done for PedersenGens: since it does not create commitments, but only verifies them.

```rust
let p = Prover::new(&pc_gens, transcript);
...
let proof = p.prove(&bp_gens)?;
```
```rust
let v = Verifier::new(transcript);
...
v.verify(proof, &pc_gens, &bp_gens)?;
```

### DIFF
--- a/src/r1cs/verifier.rs
+++ b/src/r1cs/verifier.rs
@@ -310,6 +310,13 @@ impl<'a, 'b> Verifier<'a, 'b> {
         Ok(wrapped_self.verifier)
     }
 
+    /// Returns a required capacity for BulletproofGens.
+    pub fn required_capacity(&self) -> usize {
+        // XXX this does not include multipliers allocated after randomization,
+        // which is only available inside the `verify()` method when it's too late.
+        self.num_vars
+    }
+
     /// Consume this `VerifierCS` and attempt to verify the supplied `proof`.
     pub fn verify(
         mut self,
--- a/src/r1cs/verifier.rs
+++ b/src/r1cs/verifier.rs
@@ -22,7 +22,6 @@ use transcript::TranscriptProtocol;
 /// which consumes the `Verifier` instance, samples random challenges
 /// that instantiate the randomized constraints, and verifies the proof.
 pub struct Verifier<'a, 'b> {
-    bp_gens: &'b BulletproofGens,
     pc_gens: &'b PedersenGens,
     transcript: &'a mut Transcript,
     constraints: Vec<LinearCombination>,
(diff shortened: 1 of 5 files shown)


src/r1cs/verifier.rs
```diff
@@ -317,7 +311,11 @@ impl<'a, 'b> Verifier<'a, 'b> {
     }
 
     /// Consume this `VerifierCS` and attempt to verify the supplied `proof`.
-    pub fn verify(mut self, proof: &R1CSProof) -> Result<(), R1CSError> {
+    pub fn verify(
+        mut self,
+        proof: &R1CSProof,
+        bp_gens: &'b BulletproofGens,
```
hdevalence: Do we need a named lifetime parameter here? The `&BulletproofGens` reference isn't retained anywhere, so I think it just needs to outlive the `verify` function (unlike in the old code, where we needed it to outlive the entire verifier).

hdevalence:
I think it's fine to break symmetry between the parameters for proving and verification, since there's already different workflows (the verifier doesn't synthesize a witness).

I don't know what your resolution was on whether it makes sense to have capacity methods, but this LGTM otherwise!

oleganza:
For capacity current thinking is to have an API on R1CSProof to return the required capacity based on IPP length. This way, the verifier can see upfront how "expensive" the proof is and charge the cost/decrease karma/whatever based on that.

hdevalence:
That's a cool idea for the verifier; does the prover still have to pick out their generator sizes prior to making the proof when using the 2-phase API?

oleganza:
To allow prover know the required capacity beforehand we need to use FnMut closure that’ll be called twice: after first and second phases right before we compute A/S commitments. That closure is supposed to provide a generators instance for a given capacity. Prover can then extend them if needed or fail if they don’t like the capacity.

oleganza:
I suggest we work on passing a FnMut closure to let the prover adjust capacity as a separate PR.

cathieyun:
@oleganza do you want to make / link to an issue for making a FnMut closure as described in https://github.com/dalek-cryptography/bulletproofs/pull/257#issuecomment-476429341?

PR #258, opened 2019-03-16
r1cs: optional serialization for second-phase commitments

### LINKED ISSUE #242: Optional API for challenge-based R1CS proofs

oleganza:
If there were no second phase allocations, the proof is shorter by 3x32 bytes: second-phase commitments are treated as identity points.

If there were no randomization callbacks, the proof is domain-separated from one with randomization callbacks.

Note: it is possible to use randomized constraints, yet still perform no allocations in the second phase and receive a shorter proof. In this case the two-phase domain separator is still applied.

Closes #242

Alternative: https://github.com/dalek-cryptography/bulletproofs/pull/264

### DIFF
--- a/src/r1cs/proof.rs
+++ b/src/r1cs/proof.rs
@@ -3,7 +3,7 @@
 
 use curve25519_dalek::ristretto::CompressedRistretto;
 use curve25519_dalek::scalar::Scalar;
-use curve25519_dalek::traits::{IsIdentity,Identity};
+use curve25519_dalek::traits::{Identity, IsIdentity};
 
 use errors::R1CSError;
 use inner_product_proof::InnerProductProof;
--- a/src/r1cs/proof.rs
+++ b/src/r1cs/proof.rs
@@ -3,6 +3,7 @@
 
 use curve25519_dalek::ristretto::CompressedRistretto;
 use curve25519_dalek::scalar::Scalar;
+use curve25519_dalek::traits::{IsIdentity,Identity};
 
 use errors::R1CSError;
 use inner_product_proof::InnerProductProof;
--- a/src/r1cs/proof.rs
+++ b/src/r1cs/proof.rs
@@ -65,14 +65,15 @@ pub struct R1CSProof {
 }
 
 impl R1CSProof {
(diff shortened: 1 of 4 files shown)


daira:
The method of distinguishing the two serialization types (whether the number of elements is even or odd) seems a bit funky / potentially fragile to me.

oleganza:
@daira that's Henry's idea :-) I think it's acceptably funky because the proof contents are defined by the protocol and not going to change w/o incompatible change to the protocol. And the only dynamic portion is IPP proof which is guaranteed to be even number of elements.

oleganza:
I will make an alternative PR with a byte flag in the beginning of the proof and we'll see if it looks better to everyone.

oleganza:
@daira @cathieyun @hdevalence alternative PR with a byte version: https://github.com/dalek-cryptography/bulletproofs/pull/264

oleganza:
Closed by #264

PR #259, opened 2019-03-17
add build instructions to the README

greyspectrum:
In the course of getting started, I noticed that bulletproofs won't compile with stable Rust, but instead requires nightly Rust. Although the compiler was informative enough that I was able to figure it out, it would likely make it easier for other prospective contributors if this were in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,15 @@ assert!(
 );
 # }
 ```
+## Building
+
+To compile successfully, you will need to have nightly Rust installed, rather than stable.
+
+You can install nightly Rust with rustup:
+
+```text
+rustup default nightly
+```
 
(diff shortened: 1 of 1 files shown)


hdevalence:
Thanks!

Since this is a documentation change that addresses a current problem with the main branch, I think it's fair to class this as a bug-fix and shouldn't wait for the R1CS implementation to be finished (which is when the `develop` branch will get merged). So I'm going to merge this into `main` so that it shows up on the landing page.

hdevalence:
Rebased and (manually) merged as https://github.com/dalek-cryptography/bulletproofs/commit/5f4974d947c863bf1737d7f5bda40cf5f4106964

Issue #247, opened 2019-01-18
Rename R1CSProof/R1CSError to Proof/Error to use `r1cs::` prefix instead


oleganza:
Currently I have an awkward way of using `r1cs::Verifier` and `r1cs::R1CSProof`: one is prefixed with the module name, another is prefixed in its own name. Would be good to be able to do something like this:

```
use bulletproofs::r1cs;

...

r1cs::Verifier
...
r1cs::Proof
```

hdevalence:
Hmm, how does this interact with the naming scheme we have for the rangeproofs?  Is there a difference between the `Verifier` type (which is probably used only in an R1CS context) and the `R1CSProof` type (which might be in some struct somewhere outside of an R1CS context)?

oleganza:
Seems like it's "idiomatic Rust" to have simpler type names and meaningful module names? Examples:`iter` and `mem` modules with simple one-word names like `once` and `swap`.

So even outside a "R1CS context" one may see `r1cs::Proof` or `range_proof::Proof` just like one sees `iter::once` and `mem::swap`. Does this make sense?

oleganza:
At the same time, something like `R1CSVerifier` and `R1CSProof` also work, and they also follow the same style. But having one name prefixed and another one not prefixed seems awkward to me.

PR #261, opened 2019-03-17
Rename R1CSProof/R1CSError to Proof/Error to use r1cs:: prefix instead

greyspectrum:
This is an attempt to address #247. I'm quite new to Rust, so it's entirely possible that I misunderstood the intent of the issue (this PR is a simple search and replace on the type names). If that's the case, I apologize in advance and would be happy to work on this further.

### DIFF
--- a/benches/r1cs.rs
+++ b/benches/r1cs.rs
@@ -1,6 +1,6 @@
 extern crate bulletproofs;
 use bulletproofs::r1cs::{
-    ConstraintSystem, Prover, Error, Proof, RandomizedConstraintSystem, Variable, Verifier,
+    ConstraintSystem, Error, Proof, Prover, RandomizedConstraintSystem, Variable, Verifier,
 };
 use bulletproofs::{BulletproofGens, PedersenGens};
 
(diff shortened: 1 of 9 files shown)


cathieyun:
This looks perfect to me - it looks like all instances of `r1cs::R1CSProof` and `r1cs::R1CSError` have been correctly renamed to `r1cs::Proof` and `r1cs::Error`. 

I would give this an approval right away, but I think @oleganza should also take a look since he opened the issue and might have more things to add.

Thanks for your help @greyspectrum !

hdevalence:
Right now we have `RangeProof` and `R1CSProof` for the range proofs and constraint system proofs. Since we can't change the name of the `RangeProof` without breaking stability, would it be confusing to have `RangeProof` and `Proof`? The constraint system proof type is disambiguated as `r1cs::Proof` by the module name, but only when the client code imports it that way -- I don't know if that's going to always be done, for instance in Rustdoc output.

greyspectrum:
This may be helpful, potentially:

https://github.com/rust-lang/rfcs/pull/356

...although they acknowledge that this can cause confusion for common/popular type names.

oleganza:
In the linked RFC proposal there's an interesting comment:

image

I personally prefer to not use renamed imports (except in pathological cases where otherwise things are too ugly), to make the code make sense locally w/o too much context. So if there's no more-or-less global entity "TcpSocket", then making it up in the imports line is going to add another entity to the cognitive load. So it's better to either have `TcpSocket` defined in its library this way, or if it's `tcp::Socket`, then use it as such, to make that link clear.

Now, if I am in the library author's shoes, then for me the question becomes: which one should we use (`TcpSocket` or `tcp::Socket`), assuming that I don't want to encourage imports `as TcpSocket`.

greyspectrum:
If `r1cs::Proof` is potentially ambiguous, would the best alternative be to leave things as they are or preface all of the types in `r1cs` with the module name?

```
use bulletproofs::r1cs::{
    R1CSConstraintSystem, R1CSError, R1CSProof, R1CSProver, R1CSRandomizedConstraintSystem, R1CSVariable, R1CSVerifier,
};
```
It's awfully repetitive, but isn't ambiguous or inconsistent.

PR #263, opened 2019-03-19
gens: dynamically growable BulletproofGens

vickiniu:
Adds `add_gens` for `BulletproofGens` so we can extend generators if they are not big enough

### DIFF
--- a/src/generators.rs
+++ b/src/generators.rs
@@ -70,6 +70,16 @@ impl GeneratorsChain {
             reader: shake.xof_result(),
         }
     }
+
+    /// Advances the reader n times, squeezing and discarding
+    /// the result.
+    fn fast_forward(&mut self, n: usize) -> &mut Self {
+        for _ in 0..n {
+            let mut buf = [0u8; 64];
+            self.reader.read(&mut buf);
+        }
+        self
+    }
 }
 
 impl Default for GeneratorsChain {
@@ -87,6 +97,7 @@ impl Iterator for GeneratorsChain {
 
         Some(GeneratedPoint(uniform_bytes))
     }
+
     fn size_hint(&self) -> (usize, Option<usize>) {
         (usize::max_value(), None)
     }
@@ -206,7 +217,7 @@ impl BulletproofGens {
             LittleEndian::write_u32(&mut label[1..5], party_index);
             self.G_vec[i].append(
                 &mut GeneratorsChain::new(&label)
-                    .skip(self.gens_capacity)
+                    .fast_forward(self.gens_capacity)
                     .take(new_capacity - self.gens_capacity)
                     .map(|p| p.into())
                     .collect::<Vec<RistrettoPoint>>(),
@@ -215,7 +226,7 @@ impl BulletproofGens {
             label[0] = b'H';
             self.H_vec[i].append(
                 &mut GeneratorsChain::new(&label)
-                    .skip(self.gens_capacity)
+                    .fast_forward(self.gens_capacity)
                     .take(new_capacity - self.gens_capacity)
                     .map(|p| p.into())
                     .collect::<Vec<RistrettoPoint>>(),
--- a/src/generators.rs
+++ b/src/generators.rs
@@ -57,8 +57,6 @@ struct GeneratorsChain {
     reader: Sha3XofReader,
 }
 
-struct GeneratedPoint([u8; 64]);
-
 impl GeneratorsChain {
     /// Creates a chain of generators, determined by the hash of `label`.
     fn new(label: &[u8]) -> Self {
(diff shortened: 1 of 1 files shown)


src/generators.rs
```diff
@@ -181,6 +181,32 @@ impl BulletproofGens {
         }
     }
 
+    /// Precomputes additional generators and increases the capacity.
+    pub fn add_gens(&mut self, add_gens: usize) {
```
oleganza: bikeshed: let's spell out words fully and maybe name it more like `increase_capacity(new_capacity: ...)`? This way we don't need to compute how much extra items we need - we can just set exactly how much we need. No-op if capacity is sufficient.

src/generators.rs
```diff
@@ -181,6 +181,32 @@ impl BulletproofGens {
...
+        use byteorder::{ByteOrder, LittleEndian};
+
+        for i in 0..self.party_capacity {
+            let party_index = i as u32;
+            let mut label = [b'G', 0, 0, 0, 0];
+            LittleEndian::write_u32(&mut label[1..5], party_index);
+            self.G_vec[i].append(
+                &mut GeneratorsChain::new(&label)
+                    .skip(self.gens_capacity)
```
oleganza: In the current form this `.skip` is expensive - it generates all the intermediate points. To avoid this cost we need either of two things:

1. Store `GeneratorsChain` inside each item of `G_vec/H_vec` and continue iterating from where it stopped.
2. Modify `GeneratorsChain` to return some lazily-computed points: e.g. yield `GeneratedPoint([u8;64])` that implements `Into<RistrettoPoint>`. This seems to be cleaner. Then, `skip` becomes cheap.

src/generators.rs
```diff
@@ -308,4 +350,36 @@ mod tests {
         helper(16, 2);
         helper(16, 1);
     }
+
+    #[test]
+    fn add_gens_matches() {
```
hdevalence: small nit: add gens matches what?
```suggestion
    fn resizing_small_gens_matches_creating_bigger_gens() {
```

src/generators.rs
```diff
@@ -308,4 +350,36 @@ mod tests {
...
+        let gens = BulletproofGens::new(64, 8);
+
+        let mut gen_resized = BulletproofGens::new(32, 8);
+        gen_resized.increase_capacity(64);
+
+        let helper = |n: usize, m: usize| {
```
hdevalence: Do we need to do this check for each aggregation size? Since `gens.G` and `gens.H` return iterators over all the generators for a particular aggregation size (which is a subset of the generators for the "biggest" size), isn't it sufficient to check only the biggest aggregation size?

src/generators.rs
```diff
@@ -68,6 +68,16 @@ impl GeneratorsChain {
             reader: shake.xof_result(),
         }
     }
+
+    /// Advances the reader n times, squeezing and discarding
+    /// the result.
+    fn fast_forward(&mut self, n: usize) -> &mut Self {
```
oleganza: `skip` and other iterator methods consume self and return self, not the mutable borrows. Let's follow the same style here?

```suggestion
    fn fast_forward(mut self, n: usize) -> Self {
```

hdevalence:
Looks great!

src/generators.rs
```diff
@@ -181,6 +192,37 @@ impl BulletproofGens {
...
+
+        for i in 0..self.party_capacity {
+            let party_index = i as u32;
+            let mut label = [b'G', 0, 0, 0, 0];
+            LittleEndian::write_u32(&mut label[1..5], party_index);
+            self.G_vec[i].append(
+                &mut GeneratorsChain::new(&label)
+                    .fast_forward(self.gens_capacity)
+                    .take(new_capacity - self.gens_capacity)
+                    .collect::<Vec<_>>(),
```
oleganza: if you use `.extend` instead of `.append`, you can use the iterator and avoid allocation caused by `.collect`:

https://doc.rust-lang.org/std/iter/trait.Extend.html

oleganza:
```rust
            self.G_vec[i].extend(
	                GeneratorsChain::new(&label)
	                    .fast_forward(self.gens_capacity)
	                    .take(new_capacity - self.gens_capacity)
            );
```

oleganza:
Another suggestion: we can factor out common loops for `G/H_vec` in a helper method that has parameters `skip:usize` and `take:usize`. 

`::new` would instantiate `Self{ gens_capacity: 0, party_capacity: <parties>, G_vec: <empty vec per party>, H_vec: <empty vec per party> }`, and then call that helper to increase capacity with `skip=0` and `take=<initial gens capacity>`.

`::increase_capacity` would call the same helper with appropriate `skip=old_capacity` and `take=new_capacity - old_capacity`.

Initial generation would not induce many reallocations because implementation of `Extend::extend` for `Vec` makes sure to reserve enough memory for the provided iterator, and empty Vecs (used in `::new`) do not allocate anything.

vickiniu:
updated to factor our generator computation, thanks for the suggestion @oleganza !

src/generators.rs
```diff
@@ -153,34 +153,14 @@ impl BulletproofGens {
     /// * `party_capacity` is the maximum number of parties that can
     ///    produce an aggregated proof.
     pub fn new(gens_capacity: usize, party_capacity: usize) -> Self {
-        use byteorder::{ByteOrder, LittleEndian};
-
-        BulletproofGens {
+        let mut gens = BulletproofGens {
             gens_capacity,
```
oleganza: Oh, another thought: if you set initial `gens_capacity` to 0 instead of whatever user asked for, then you can avoid adding a helper `compute_generators` and instead directly call `increase_capacity()` with the intended capacity.

oleganza: sorry for my misleading suggestion earlier with skip/take arguments.

PR #264, opened 2019-03-21
r1cs: optional serialization for second-phase commitments with a version tag

### LINKED ISSUE #242: Optional API for challenge-based R1CS proofs

oleganza:
If there were no second phase allocations, the proof is shorter by 3x32 bytes: second-phase commitments are treated as identity points.

If there were no randomization callbacks, the proof is domain-separated from one with randomization callbacks.

Note: it is possible to use randomized constraints, yet still perform no allocations in the second phase and receive a shorter proof. In this case the two-phase domain separator is still applied.

Closes #242
Closes #258 

Prior iteration: #258, diff: https://github.com/dalek-cryptography/bulletproofs/compare/oleg/optional-2phase...oleg/optional-2phase-tagged

### DIFF
--- a/src/r1cs/proof.rs
+++ b/src/r1cs/proof.rs
@@ -12,6 +12,9 @@ use util;
 use serde::de::Visitor;
 use serde::{self, Deserialize, Deserializer, Serialize, Serializer};
 
+const ONE_PHASE_COMMITMENTS: u8 = 0;
+const TWO_PHASE_COMMITMENTS: u8 = 1;
+
 /// A proof of some statement specified by a
 /// [`ConstraintSystem`](::r1cs::ConstraintSystem).
 ///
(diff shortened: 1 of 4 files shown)


oleganza:
here's a diff between this PR and #258:

https://github.com/dalek-cryptography/bulletproofs/compare/oleg/optional-2phase...oleg/optional-2phase-tagged

PR #266, opened 2019-03-26
bulletproofs: add closure to increase prover generator capacity

### LINKED ISSUE #265: r1cs: add closure for prover to adjust capacity
Add a `FnMut` closure to `Prover::prove(...)` to allow the prover to adjust the generator capacity based on the circuit size for the given constraint system.

See: https://github.com/dalek-cryptography/bulletproofs/pull/257#issuecomment-476429341

vickiniu:
Fixes #265

### DIFF
--- a/benches/r1cs.rs
+++ b/benches/r1cs.rs
@@ -122,7 +122,7 @@ impl KShuffleGadget {
 
     pub fn prove<'a, 'b>(
         pc_gens: &'b PedersenGens,
-        bp_gens: &'b BulletproofGens,
+        bp_gens: &'b mut BulletproofGens,
         transcript: &'a mut Transcript,
         input: &[Scalar],
         output: &[Scalar],
@@ -156,7 +156,7 @@ impl KShuffleGadget {
             .unzip();
 
         Self::fill_cs(&mut prover, input_vars, output_vars)?;
-        let proof = prover.prove(&bp_gens)?;
+        let proof = prover.prove(bp_gens, |capacity, gens| gens.increase_capacity(capacity))?;
 
         Ok((proof, input_commitments, output_commitments))
     }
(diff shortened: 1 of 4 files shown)


tests/r1cs.rs
```diff
@@ -447,3 +459,56 @@ fn range_proof_helper(v_val: u64, n: usize) -> Result<(), R1CSError> {
...
+        prover_capacity_resize_helper(0, |capacity, gens| gens.increase_capacity(capacity)).is_ok()
+    );
+
+    let (proof, com) =
+        prover_capacity_resize_helper(64, |capacity, gens| gens.increase_capacity(capacity))
+            .unwrap();
+
+    // Construct verifier generators
+    let pc_gens = PedersenGens::default();
+    let mut bp_gens = BulletproofGens::new(128, 1);
```
vickiniu: Should there be an interface for the verifier to get the needed capacity for a proof (an added field in R1CSProof)?

benches/r1cs.rs
```diff
@@ -156,7 +156,7 @@ impl KShuffleGadget {
             .unzip();
 
         Self::fill_cs(&mut prover, input_vars, output_vars)?;
-        let proof = prover.prove(&bp_gens)?;
+        let proof = prover.prove(bp_gens, |capacity, gens| gens.increase_capacity(capacity))?;
```
oleganza: maybe instead of sending generators into `prove`, and then getting them into closure, instead simply provide the reference from the closure (not sure if borrowck would love this):

```rust
let gens = BulletproofGens::new(128, 1);
...
let proof = prover.prove(|capacity| gens.increase_capacity(capacity); &gens )?;
```

?

oleganza: And the closure must return `Result<&BulletproofGens, R1CSError>` because the prover may decide to return Err if capacity is too high instead of blinding increasing it.

vickiniu: Having the closure provide the reference makes the borrow checker unhappy 😢 says it can't infer the appropriate lifetime?

vickiniu: The prover can provide a closure that doesn't resize the generators (which would result in a failure from insufficient capacity). Do we want to also have the closure return `Result<(), R1CSError>`? It feels maybe unnecessary to me since a no-op closure will have the same overall effect

PR #267, opened 2019-03-29
fix identity check with CompressedRistretto

lovesh:
Signed-off-by: lovesh harchandani <lovesh.bond@gmail.com>

### DIFF
--- a/src/r1cs/proof.rs
+++ b/src/r1cs/proof.rs
@@ -122,7 +122,7 @@ impl R1CSProof {
 
     fn missing_phase2_commitments(&self) -> bool {
         let identity = CompressedRistretto::identity();
-        self.A_I2 == identity && self.A_O2  == identity && self.S2  == identity
+        self.A_I2 == identity && self.A_O2 == identity && self.S2 == identity
     }
 
     /// Deserializes the proof from a byte slice.
--- a/src/r1cs/proof.rs
+++ b/src/r1cs/proof.rs
@@ -121,7 +121,8 @@ impl R1CSProof {
     }
 
     fn missing_phase2_commitments(&self) -> bool {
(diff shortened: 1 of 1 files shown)


isislovecruft:
Hi @lovesh, thanks for the PR! I'm not sure I see what this fixes.. could you elaborate more what the problem is?

lovesh:
Hi @isislovecruft, the trait `IsIdentity` is not implemented for `CompressedRistretto`. Hence, when i run `cargo +nightly check --all-features` on develop branch i get this error 
```
error[E0599]: no method named `is_identity` found for type `curve25519_dalek::ristretto::CompressedRistretto` in the current scope
   --> src/r1cs/proof.rs:124:19
    |
124 |         self.A_I2.is_identity() && self.A_O2.is_identity() && self.S2.is_identity()
    |                   ^^^^^^^^^^^
    |
    = note: the method `is_identity` exists but the following trait bounds were not satisfied:
            `curve25519_dalek::ristretto::CompressedRistretto : curve25519_dalek::traits::IsIdentity`
``` 
Now i could either implement `IsIdentity` for `CompressedRistretto` or do the equality comparison with the identity element. I went with the former.

hdevalence:
Hi @lovesh, try running `cargo update`? Thanks for giving some more context on the problem you ran into.

There was a minor version update to `curve25519-dalek` that added that trait impl.  The `bulletproofs`  `Cargo.toml` should have specified `1.0.3` instead of `1`, but by mistake this wasn't done in the `develop` branch (because when the code was added, Cargo resolved the dependency to a compatible version).  This problem was fixed in the `main` branch and the fix will get ported to the `develop` branch soon.  In the meantime, running `cargo update` should push the dependency to a compatible version.

lovesh:
Hi @hdevalence, thanks for the suggestion. It works now.

PR #269, opened 2019-04-23
r1cs: lease transcript from CS to bind proof to the data made available later

oleganza:
This adds method `transcript()` to the `ConstraintSystem` trait. The motivation is to allow higher-level protocol to bind the R1CS proof to data that's made available _during_ CS construction, but _not before_.

Example: in [ZkVM](https://github.com/interstellar/slingshot/blob/5377925ac10b13e0f6c958c4d037e13504cc21ce/zkvm/src/prover.rs#L108-L114) the transaction ID is computed via program execution that's also constructing a CS state. By the end of the execution, both the txid and set of constraints are known, and it's safer to bind the proof to the txid, but the transcript is exclusively held by the CS prover. This API allows to borrow the transcript from the CS prover to commit txid, and then complete computation of the R1CS proof.

### DIFF
--- a/src/r1cs/constraint_system.rs
+++ b/src/r1cs/constraint_system.rs
@@ -21,7 +21,7 @@ pub trait ConstraintSystem {
     type RandomizedCS: RandomizedConstraintSystem;
 
     /// Leases the proof transcript to the user, so they can
-    /// add extra data to which the proof must be bound, but that
+    /// add extra data to which the proof must be bound, but which
     /// is not available before creation of the constraint system.
     fn transcript(&mut self) -> &mut Transcript;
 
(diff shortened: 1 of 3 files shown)


src/r1cs/constraint_system.rs
```diff
@@ -19,6 +20,11 @@ pub trait ConstraintSystem {
     /// Represents a concrete type for the CS in a randomization phase.
     type RandomizedCS: RandomizedConstraintSystem;
 
+    /// Leases the proof transcript to the user, so they can
+    /// add extra data to which the proof must be bound, but that
```
oleganza:
```suggestion
    /// add extra data to which the proof must be bound, but which
```

PR #270, opened 2019-04-23
r1cs: decouple randomization from CS

### LINKED ISSUE #268: r1cs: prohibit nested CS randomization calls
After talking with Benedikt and Cathie a couple of weeks ago in Stanford, we found out that it is quite plausible to allow multi-phase challenges in R1CS API. To facilitate doing this in the future, but also making the current API cleaner, I propose that we disallow calling `specify_randomized_constraints()` in the `RandomizedConstraintSystem` (currently it immediately invokes the callback). In the future we might need to enable this to support deeper-phase randomization.

1. Definition of `specify_randomized_constraints()` is moved out of the `ConstraintSystem` trait into a new trait `RandomizableConstraintSystem`.
2. Trait `RandomizedConstraintSystem` inherits `ConstraintSystem`, but not the `RandomizableConstraintSystem`.
3. Gadgets that do not use randomization can use trait bound `<CS: ConstraintSystem>` as usual.
4. Gadgets like [`cloak`](https://github.com/interstellar/slingshot/blob/main/spacesuit/src/cloak.rs) or [`value_shuffle`](https://github.com/interstellar/slingshot/blob/main/spacesuit/src/shuffle.rs#L54) in Spacesuit that need to use randomization would explicitly opt-in the randomization via `<CS: ConstraintSystem+RandomizableConstraintSystem>`.
5. To allow nesting within an already opened randomization scope (like `scalar_shuffle` within `value_shuffle`), the gadget may instead choose to make a bound `<CS: RandomizedConstraintSystem>`. Then this gadget would have to be used within randomization closure opened by the user, and also guarantee that all variables it allocates are fully determined by previously committed or determined variables (which is the case with composition of two shuffle gadgets).

In the future, if we decide to support N-phase CS API, we could add `RandomizableConstraintSystem` trait to the `RandomizedConstraintSystem` and implement it appropriately by creating additional phases for nested calls to `specify_randomized_constraints`.

oleganza:
Closes #268.

### DIFF
--- a/benches/r1cs.rs
+++ b/benches/r1cs.rs
@@ -73,7 +73,7 @@ For K = 1:
 struct KShuffleGadget {}
 
 impl KShuffleGadget {
-    fn fill_cs<CS: ConstraintSystem>(
+    fn fill_cs<CS: RandomizableConstraintSystem>(
         cs: &mut CS,
         x: Vec<Variable>,
         y: Vec<Variable>,
--- a/docs/r1cs-docs-example.md
+++ b/docs/r1cs-docs-example.md
@@ -75,7 +75,7 @@ use rand::thread_rng;
 struct ShuffleProof(R1CSProof);
 
 impl ShuffleProof {
-    fn gadget<CS: ConstraintSystem>(cs: &mut CS, x: Vec<Variable>, y: Vec<Variable>) -> Result<(),R1CSError> {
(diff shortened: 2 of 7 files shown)


oleganza:
Before this is ready for review, I will prepare a PR on [Slingshot](https://github.com/interstellar/slingshot/) that uses this patch to see how it works.

oleganza:
Here's the PR against Spacesuit+ZkVM that uses the updated API: https://github.com/interstellar/slingshot/pull/291

Issue #244, opened 2019-01-17
R1CS: Closure of challenge phase is `Fn`.


rubdos:
Since #222, one writes the challenge phase using a closure. This closure gets stored in a `Vec<Box<Fn(...)>>`, and then gets called once at the start of the second phase.

Ideally, this closure should be `FnOnce`, such that it's able to consume some of its context (e.g. a set of variables could be consumed instead of borrowed).

Note that it is **currently impossible** to call a `Box<FnOnce>`, because it would call a method with unsized self. Cfr. https://github.com/rust-lang/rust/issues/28796

Nightly does support `FnBox()`, which is a temporary workaround. The above issue is discussion deprecation of `FnBox()` in favour of the `unsized_locals` feature (which makes `Box<FnOnce>` callable).

---

Note that there does not seem to be a fix for stable Rust, but I thought this should be noted somewhere. I'm not sure whether you are in favour of keeping this issue open while waiting for [`unsized_locals` stabilization](https://github.com/rust-lang/rust/issues/48055).

oleganza:
Thanks for opening this issue and explaining the context behind `Box<Fn>`. I think it makes total sense to keep it open until there is a fix for `Box<FnOnce>`.

oleganza:
Since we have to use nightly for other reasons (non-lexical lifetimes in IPP `nll`), and that `Box<FnOnce>` is getting [stabilized in Rust 1.35](https://github.com/rust-lang/rust/issues/28796#issuecomment-481502369) (next release), we could probably change the API now and throw in a `#![feature(unsized_locals)]` until 1.35.

cathieyun:
That sounds good to me! As long as we remember to remove `unsized_locals` when 1.35 comes out so we have fewer nightly dependencies.

hdevalence:
+1 to all of the above. One thing we could do is to make the `unsized_locals` only enabled when the `yoloproofs` (which can be renamed to `r1cs` or something later if we decide it's not yolo), so that it's only pulled in for the r1cs code rather than for the rangeproof code.  It might also be good to check on the NLL status.

oleganza:
Seems like NLL is stable since 1.31: https://blog.rust-lang.org/2018/12/06/Rust-1.31-and-rust-2018.html#non-lexical-lifetimes

rubdos:
Closed through #276

PR #276, opened 2019-05-28
Make closure of challenge phase FnOnce.

rubdos:
It allows for the challenge phase to accept closures that take some more
context, e.g. moving non-Clone values in the challenge closure.

This works in stable since Rust 1.35; cfr.
https://github.com/rust-lang/rust/issues/28796.

This closes issue #244.

---

~~In other news: I'm not yet happy with the `'static` bound on the closures. Since I'm not yet worried about that kind of performance (just hacking around currently), I'm not going to fight the borrow checker there, yet.~~ Cfr #323

### DIFF
--- a/src/r1cs/constraint_system.rs
+++ b/src/r1cs/constraint_system.rs
@@ -106,7 +106,7 @@ pub trait RandomizableConstraintSystem: ConstraintSystem {
     /// ```
     fn specify_randomized_constraints<F>(&mut self, callback: F) -> Result<(), R1CSError>
     where
-        F: 'static + Fn(&mut Self::RandomizedCS) -> Result<(), R1CSError>;
+        F: 'static + FnOnce(&mut Self::RandomizedCS) -> Result<(), R1CSError>;
 }
 
 /// Represents a constraint system in the second phase:
--- a/src/r1cs/prover.rs
+++ b/src/r1cs/prover.rs
@@ -37,7 +37,8 @@ pub struct Prover<'g, T: BorrowMut<Transcript>> {
 
     /// This list holds closures that will be called in the second phase of the protocol,
     /// when non-randomized variables are committed.
-    deferred_constraints: Vec<Box<dyn Fn(&mut RandomizingProver<'g, T>) -> Result<(), R1CSError>>>,
+    deferred_constraints:
+        Vec<Box<dyn FnOnce(&mut RandomizingProver<'g, T>) -> Result<(), R1CSError>>>,
 
     /// Index of a pending multiplier that's not fully assigned yet.
     pending_multiplier: Option<usize>,
@@ -182,7 +183,7 @@ impl<'g, T: BorrowMut<Transcript>> RandomizableConstraintSystem for Prover<'g, T
 
     fn specify_randomized_constraints<F>(&mut self, callback: F) -> Result<(), R1CSError>
     where
-        F: 'static + Fn(&mut Self::RandomizedCS) -> Result<(), R1CSError>,
+        F: 'static + FnOnce(&mut Self::RandomizedCS) -> Result<(), R1CSError>,
     {
         self.deferred_constraints.push(Box::new(callback));
         Ok(())
--- a/src/r1cs/verifier.rs
+++ b/src/r1cs/verifier.rs
@@ -43,7 +43,8 @@ pub struct Verifier<T: BorrowMut<Transcript>> {
     /// when non-randomized variables are committed.
     /// After that, the option will flip to None and additional calls to `randomize_constraints`
     /// will invoke closures immediately.
-    deferred_constraints: Vec<Box<dyn Fn(&mut RandomizingVerifier<T>) -> Result<(), R1CSError>>>,
+    deferred_constraints:
+        Vec<Box<dyn FnOnce(&mut RandomizingVerifier<T>) -> Result<(), R1CSError>>>,
 
     /// Index of a pending multiplier that's not fully assigned yet.
     pending_multiplier: Option<usize>,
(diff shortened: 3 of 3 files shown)


rubdos:
Seems like the CI fails on the Rust version (which should be 1.35) for this to work.

rubdos:
The new build failure is because of the bunch of recent deprecations in merlin.

hdevalence:
Hi, those deprecations were fixed in the `main` branch of this crate but those changes weren't merged into the `develop` branch. I can update the `develop` branch so that this PR can work again.

rubdos:
Is `develop` still being used then? Otherwise I'll rebase this on `main`.

hdevalence:
Yep, it's still being used, it's just that those changes were included as part of a PR that added additional validation to the rangeproof code, so they were put into the `main` branch.  The `develop` branch is still for ongoing work on the R1CS implementation, which slowed a bit but will probably pick up again soon.

rubdos:
Ok, fair! So, would you mind porting those deprecation fixes to develop then, then I'll rebase.

hdevalence:
Currently in-progress with #277!

rubdos:
I've rebased this over `develop`. I'm looking into the `'static` lifetime too now, so hold your horses for a few minutes :-)

rubdos:
Looks like I managed to shorten the lifetime. I added a test that would fail before and succeed after the patch. It does however mean that all `Prover`s and `Verifier`s, and `RandomizableConstraintSystem` all get some extra lifetimes (which technically would mean breaking the API).  Since it's behind a yolo-feature, I assume that's not a big issue.

oleganza:
Does non-static lifetime work well with multiple calls to `specify_randomized_constraints `? I'll try this out with `zkvm` and `spacesuit` next week.

oleganza:
That's an exciting update, thanks!

rubdos:
> Does non-static lifetime work well with multiple calls to `specify_randomized_constraints `? I'll try this out with `zkvm` and `spacesuit` next week.

It should! I currently have no use for it myself, but I may soon have. If it doesn't work out, feel free to add a test case and I'll take a look.

src/r1cs/constraint_system.rs
```diff
@@ -104,9 +104,12 @@ pub trait RandomizableConstraintSystem: ConstraintSystem {
     ///     // ...
     /// })
     /// ```
-    fn specify_randomized_constraints<F>(&mut self, callback: F) -> Result<(), R1CSError>
+    fn specify_randomized_constraints<'a: 'constraints, F>(
+        &mut self,
+        callback: F,
+    ) -> Result<(), R1CSError>
```
rubdos, line 110: I'm actually unsure whether the additional `'a` lifetime is needed here. IIRC, any closure `FnOnce + 'constraints` should also implement `FnOnce + 'a`, since `'a: 'constraints`, right? I've never had to do a lot of lifetime stuff, but I think the announced experiment of @oleganza should point out anything strange here.

tests/r1cs.rs
```diff
@@ -452,3 +452,38 @@ fn range_proof_helper(v_val: u64, n: usize) -> Result<(), R1CSError> {
...
+    let mut verifier_transcript = Transcript::new(b"NonStaticRandomClosuresTest");
+    let mut verifier = Verifier::new(&mut verifier_transcript);
+    verifier
+        .specify_randomized_constraints(|rcs| {
+            rcs.constrain((Scalar::from(a.clone()) - Scalar::one()).into());
+            Ok(())
+        })
+        .unwrap();
+    verifier.verify(&proof, &pc_gens, &bp_gens).unwrap();
+}
```
rubdos, line 489: Maybe this needs an additional interesting test-case for multiple different lifetimes and borrows?

src/r1cs/constraint_system.rs
```diff
@@ -104,9 +104,12 @@ pub trait RandomizableConstraintSystem: ConstraintSystem {
     ///     // ...
     /// })
     /// ```
-    fn specify_randomized_constraints<F>(&mut self, callback: F) -> Result<(), R1CSError>
+    fn specify_randomized_constraints<'a: 'constraints, F>(
+        &mut self,
```
rubdos, line 108: Not specifying an explicit lifetime on `&mut self` might get people into trouble [with late-bound lifetimes](https://github.com/rust-lang/rust/issues/42868). I've had that while experimenting on it, but haven't come to fix it yet. I think it should be possible to add it later though, without breaking API...

src/r1cs/constraint_system.rs
oleganza: first thing i noticed is that now every gadget has to specify additional lifetime parameters

image

rubdos: I'm not sure whether it's possible to mitigate tbqh, but I'm having a look into this. Quite the opportunity to learn a ton about lifetimes for me!

rubdos: I'm actually quite confident now that it's not possible without adding a lifetime parameter to the `RandomizableConstraintSystem`: for a given constraint system `let p = Prover::new(...)`, we need to declare the lifetime of the `FnOnce`s that are stored in the system. Prover can only implement `RandomizableConstraintSystem` when the randomized constraint you are adding outlives that lifetime of the constraintsystem.

Think of it like "Prover only implements RandomizableConstraintSystem for constraints that reference things that outlive the Prover."

... it is indeed quite unfortunate that this has to break the API though, for gadgets that use the RCS.

oleganza: How about upgrading from Fn to FnOnce in this PR, but keeping the 'static requirement, and then separately explore if moving to generic lifetime bound is worth it? Right now i have to propagate 'a in a ton of places in ZkVM, because where I had a `vm::Verifier`, now I have `vm::Verified<'a>`, and it leaks to `PrecomputedTx` that wraps it, etc. I could probably slap 'static at some point inside ZkVM/Spacesuit, but that push the burden of justification on my codebase and I'm just... ugh... not ready for it right now :-P

rubdos:
I dropped the non-`'static` stuff from this PR and created #323 in favour of that. I'll rebase this on develop now.

tests/r1cs.rs
rubdos: Dismissed to #323

src/r1cs/constraint_system.rs
rubdos: Dismissed to #323.

PR #277, opened 2019-06-04
Merge main into develop

hdevalence:
This updates the develop branch with R1CS code to include the extra validation changes made in the main branch for 1.0.2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 Entries are listed in reverse chronological order.
 
+## 1.0.2
+
+* Updates the library to use the renamed functions in Merlin 1.1.
+* Adds additional validation checks to prevent identity points being used as
+  part of a proof.  This does not appear to have security content, but is
+  intended as a defense-in-depth mechanism.  
+  See [this comment][identity_comment] for more motivation.
+* Documentation tweaks.
+
 ## 1.0.1
 
 * Tweaks to crate metadata.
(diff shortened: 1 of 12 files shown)


hdevalence:
Ah, looks like this doesn't quite work, because of the lack of error conversions...

hdevalence:
Updated.  There's an issue with the R1CS benchmarks -- it looks like they fell out of sync with the `RandomizedConstraintSystem` / `RandomizableConstraintSystem` change.

In the long term, I think that it would be better to combine R1CSError with ProofError, if we can do that without breaking stability.  If not, we should refactor the TranscriptProtocol so that it returns a TranscriptError type that can be converted into either a ProofError or an R1CSError.  For now I would prefer to keep the "bad conversion" and then remove it later by one of those two means.

rubdos:
Have you considered using something like [`failure`](https://github.com/rust-lang-nursery/failure) for the exported error types? In my experience it aids with three things:

1. Rapid writing of new errors using `bail!()` et al., if you're prototyping something new.
2. Easily returning "generic" errors using the `Error` struct and `?` ...
3. ... while it still allows to return non-generic errors (if performance might be at stake).

hdevalence:
Yep, we are using `failure`, but we don't want to use `Error` because it's not as lightweight as we'd like. I think that the correct fix is to add a `TranscriptError` type.

hdevalence:
We should probably update and merge this branch before merging #280.  I think the things left to do here are:

- [ ] Resolve the error-heirarchy issue (probably by creating a new `TranscriptError` that can be converted to a `ProofError` or an `R1CSError`)
- [ ] Fix the R1CS benchmarks

hdevalence:
Fixed the R1CS benchmarks.

hdevalence:
Moved the error item into #286, let's fix it there instead of here.

Issue #275, opened 2019-05-23
`no_std` API


isislovecruft:
Hi! It was briefly discussed in the dalek-cryptography slack that this crate currently cannot be used in `no_std` environments due to its internal usage of `rand::thread_rng()`, which obviously requires thread-local storage and thus the `std` library. @oleganza proposed exposing an API such as `verify_with_rng()`, which then the current `verify()` would be a backwards-compatible wrapper for by passing in a `rand::thread_rng()`. `verify_with_rng()` should have a trait bound for `Rng + CryptoRng` (and possibly also `SeedableRng`? since the `FromEntropy` trait also requires `std`).

Questions:

1. Would you like me to feature-gate the publicity of this? I.e.:
  ```
  #[cfg(not(feature = "std"))]
  pub fn verify_with_rng() { ... }

  #[cfg(feature = "std")]
  fn verify_with_rng() { ... }
  ```
  So that explicitly only `no_std` users are trusted with passing in their own correctly-seeded PRNG?

oleganza:
Doesn't this equally apply to `prove/prove_with_rng`?

oleganza:
I'd vote for not depriving std-users from having `prove_with_rng`. It's useful to compose with higher-level APIs that make the same distinction for the same reason and extra restrictions would force higher-level libs to make exactly the same dichotomy as lower-level lib.

As of footguns, i think it's enough that the safer-to-use API is (1) shorter-named (2) takes fewer arguments.

hdevalence:
I would prefer not to have a `verify_with_rng` and instead find a way for verification be a stand-alone operation that depends only on the proof.

oleganza:
Unfortunately, verification involves merging multiple statements in one with a random factor. Here are the available options to make `thread_rng` dependency optional or remove entirely:

1. `verify` / `verify_with_rng` as mentioned above.
2. `verify` uses deterministic challenges w/o any RNG at all (via Transcript instance). This is a bit weaker since attacker can learn those, but it seems to be strictly the same trick as deterministic merge of statements within the previous steps of the protocols, so if that's a weakness, the whole protocol is weak already.
3. `verify` uses internally `thread_rng` if built with `std` and uses deterministic challenges if not. That's IMHO a poor dichotomy akin to "try /dev/random, and if it's blocking, try /dev/urandom instead". This makes other people think that one option is clearly worse, but somehow still acceptable.

PR #280, opened 2019-07-22
Implement no_std support

xoloki:
This change implements no_std support for the bulletproofs crate.  The changes are mostly mechanical.  This is the culmination several PRs worth of work, with changes to curve25519 and merlin crates already merged and released.

It was necessary to make minor modifications to some APIs, particularly in regards to RNGs.  As discussed by @isislovecruft in #275, the APIs in this change are backwards compatible for those building the crate in std mode.  So current users of the bulletproofs crate will see no change.  Users wishing to build bulletproofs in no_std mode will need to call the new *_with_rng prove and verfiy functions, and pass in a RNG.

This change is based on the main branch rather than the develop branch.  This is because the main branch in the dalek repo has several changes that are not in the develop branch.  This made rebasing my change on the upstream develop branch to be problematic.

### DIFF
--- a/src/range_proof/dealer.rs
+++ b/src/range_proof/dealer.rs
@@ -14,6 +14,10 @@ cfg_if::cfg_if! {
     }
 }
 
+<<<<<<< HEAD
+=======
+use curve25519_dalek::ristretto::RistrettoPoint;
+>>>>>>> use naive curve point serialization in internal data structures
 use curve25519_dalek::scalar::Scalar;
 use merlin::Transcript;
 
@@ -105,18 +109,19 @@ impl<'a, 'b> DealerAwaitingBitCommitments<'a, 'b> {
 
         // Commit each V_j individually
         for vc in bit_commitments.iter() {
-            self.transcript.append_point(b"V", &CompressedRistretto::from_slice(&vc.V_j));
+            self.transcript.append_point(b"V", &vc.V_j);
         }
 
-        // Append aggregated A_j, S_j
-        let A: RistrettoPoint = bit_commitments.iter().map(|vc| CompressedRistretto::from_slice(&vc.A_j).decompress().unwrap()).sum();
+        // Commit aggregated A_j, S_j
+        let A: RistrettoPoint = bit_commitments.iter().map(|vc| vc.A_j).sum();
         self.transcript.append_point(b"A", &A.compress());
 
-        let S: RistrettoPoint = bit_commitments.iter().map(|vc| CompressedRistretto::from_slice(&vc.S_j).decompress().unwrap()).sum();
+        let S: RistrettoPoint = bit_commitments.iter().map(|vc| vc.S_j).sum();
         self.transcript.append_point(b"S", &S.compress());
+>>>>>>> use naive curve point serialization in internal data structures
 
-        let y = self.transcript.challenge_scalar(b"y").to_bytes();
-        let z = self.transcript.challenge_scalar(b"z").to_bytes();
+        let y = self.transcript.challenge_scalar(b"y");
+        let z = self.transcript.challenge_scalar(b"z");
         let bit_challenge = BitChallenge { y, z };
 
         Ok((
@@ -166,13 +171,13 @@ impl<'a, 'b> DealerAwaitingPolyCommitments<'a, 'b> {
         }
 
         // Commit sums of T_1_j's and T_2_j's
-        let T_1: RistrettoPoint = poly_commitments.iter().map(|pc| CompressedRistretto::from_slice(&pc.T_1_j).decompress().unwrap()).sum();
-        let T_2: RistrettoPoint = poly_commitments.iter().map(|pc| CompressedRistretto::from_slice(&pc.T_2_j).decompress().unwrap()).sum();
+        let T_1: RistrettoPoint = poly_commitments.iter().map(|pc| pc.T_1_j).sum();
+        let T_2: RistrettoPoint = poly_commitments.iter().map(|pc| pc.T_2_j).sum();
 
         self.transcript.append_point(b"T_1", &T_1.compress());
         self.transcript.append_point(b"T_2", &T_2.compress());
 
-        let x = self.transcript.challenge_scalar(b"x").to_bytes();
+        let x = self.transcript.challenge_scalar(b"x");
         let poly_challenge = PolyChallenge { x };
 
         Ok((
@@ -228,9 +233,9 @@ impl<'a, 'b> DealerAwaitingProofShares<'a, 'b> {
             return Err(MPCError::WrongNumProofShares);
         }
 
-        let t_x: Scalar = proof_shares.iter().map(|ps| Scalar::from_bytes_mod_order(ps.t_x)).sum();
-        let t_x_blinding: Scalar = proof_shares.iter().map(|ps| Scalar::from_bytes_mod_order(ps.t_x_blinding)).sum();
-        let e_blinding: Scalar = proof_shares.iter().map(|ps| Scalar::from_bytes_mod_order(ps.e_blinding)).sum();
+        let t_x: Scalar = proof_shares.iter().map(|ps| ps.t_x).sum();
+        let t_x_blinding: Scalar = proof_shares.iter().map(|ps| ps.t_x_blinding).sum();
+        let e_blinding: Scalar = proof_shares.iter().map(|ps| ps.e_blinding).sum();
 
         self.transcript.append_scalar(b"t_x", &t_x);
         self.transcript
@@ -246,21 +251,13 @@ impl<'a, 'b> DealerAwaitingProofShares<'a, 'b> {
             .take(self.n * self.m)
             .collect();
 
-        let l_vec_bytes: Vec<CurveScalar> = proof_shares
+        let l_vec: Vec<Scalar> = proof_shares
             .iter()
             .flat_map(|ps| ps.l_vec.clone().into_iter())
             .collect();
-        let l_vec: Vec<Scalar> = l_vec_bytes
-            .iter()
-            .map(|ps| Scalar::from_bytes_mod_order(*ps))
-            .collect();
-        let r_vec_bytes: Vec<CurveScalar> = proof_shares
+        let r_vec: Vec<Scalar> = proof_shares
             .iter()
             .flat_map(|ps| ps.r_vec.clone().into_iter())
-            .collect();
-        let r_vec: Vec<Scalar> = r_vec_bytes
-            .iter()
-            .map(|ps| Scalar::from_bytes_mod_order(*ps))
             .collect();
 
         let ipp_proof = inner_product_proof::InnerProductProof::create(
@@ -306,7 +303,7 @@ impl<'a, 'b> DealerAwaitingProofShares<'a, 'b> {
     ) -> Result<RangeProof, MPCError> {
         let proof = self.assemble_shares(proof_shares)?;
 
-        let Vs: Vec<_> = self.bit_commitments.iter().map(|vc| CompressedRistretto::from_slice(&vc.V_j)).collect();
+        let Vs: Vec<_> = self.bit_commitments.iter().map(|vc| vc.V_j).collect();
 
         // See comment in `Dealer::new` for why we use `initial_transcript`
         let transcript = &mut self.initial_transcript;
--- a/src/range_proof/dealer.rs
+++ b/src/range_proof/dealer.rs
@@ -6,12 +6,11 @@
 
 use core::iter;
 
-cfg_if::cfg_if! {
-    if #[cfg(feature = "alloc")] {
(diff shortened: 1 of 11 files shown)


oleganza:
This looks cool. Will check in more detail tomorrow.

hdevalence:
It looks like one of the CI failures is just because of rustfmt changes that haven't been applied to the codebase yet; I can make a separate change that fixes them so that this branch can be merged cleanly on top of the fixed style.

arthurgreef:
Hi - I pulled the branch but I cannot bet a clean build. Can you provide me with you build command? Is is cargo build --no-default-features --features alloc? Thanks.

xoloki:
> Hi - I pulled the branch but I cannot bet a clean build. Can you provide me with you build command? Is is cargo build --no-default-features --features alloc? Thanks. 

I pushed a change to the branch that forces the rand dependency to 0.6, which appears to fix this.  Your cargo build command will now work.  

I was testing the no_std support in an application that used bulletproofs in no_std mode, so it is likely that some other dependency was already requesting 0.6 and cargo's dependency resolution was never actually using a 0.5 rand.

xoloki:
> It looks like one of the CI failures is just because of rustfmt changes that haven't been applied to the codebase yet; I can make a separate change that fixes them so that this branch can be merged cleanly on top of the fixed style.

The badly formatted code was all in places that I changed, so I went ahead and fixed my branch @hdevalence

xoloki:
All checks are now passing.

arthurgreef:
We are using  cargo-nono to see if the libraries reference std and are noticing that the dependencies do. I that expected? Thanks.

hdevalence:
Re: `curve25519-dalek` failure, it might be because the version with `no_std` fixes hasn't been released yet, I can check and do a new release if not. There's a security audit soon-to-be-released for these libraries with some minor issues so we might batch `no_std` fixes into those fixes instead of doing two at once.

I updated the `develop` branch so there should not be any more divergence between the `main` and `develop` branches.

hdevalence:
Hi all, recent changes to Rust nightlies broke the build (by changing the path root for doc(include) path specifiers), so we had to update the `develop` branch and release `1.0.4` which works on current nightlies.  The current `develop` branch is now back in sync with the `main` branch, is ready for more changes to be merged in, and won't have any more drastic changes (except to the R1CS code).

xoloki:
Okay @hdevalence I rebased on upstream/develop, force pushed the branch, and updated the PR to point to develop.  Let me know if you want me to do anything else.

xoloki:
@arthurgreef cargo-nono operates by parsing the source, then analyzing the AST tokens to look for stuff that seems problematic.  It is not particularly reliable, and frequently returns false positives.  For my merlin PR I actually ended up patching cargo-nono to get it to stop returning false positives.  

In this case, it appears that cargo-nono is not properly identifying how curve25519 is using preprocessor directives to selectively enable the use std:: statements.  This seems to be a result of the multi-leveled modular structure of the curve25519 crate.  

While I was able to easily patch cargo-nono for the merlin PR, there appears to be no way to easily patch it for this PR to remove the false positive on curve25519.

xoloki:
All tests are now passing @hdevalence.

hdevalence:
Great! Making the R1CS backend depend on `std` is fine, it's less clearly useful in an embedded context and the code is under a lot more change than the rangeproof code.

Instead of adding a dependency on `cfg_if`, it's possible to annotate the import, module, etc. statements directly, e.g., instead of 

```
cfg_if::cfg_if! {
    if #[cfg(feature = "alloc")] {
        extern crate alloc;
        use alloc::vec::Vec;
    }
}
```
the attributes can be added to the particular statements:
```
#[cfg(feature = "alloc")]
extern crate alloc;
#[cfg(feature = "alloc")]
use alloc::vec::Vec;
```

I think that all of the config statements can be handled this way, I don't think there's a need for `cfg_if`.

burdges:
Just fyi, I noticed this does not work
```
#[cfg(feature = "alloc")]
if foo() { bar() }
```
while this does work
```
#[cfg(feature = "alloc")]
{ if foo() { bar() } }
```
I'm not sure why, but also not really sure what good cfg_if does.

xoloki:
Ok @hdevalence I removed cfg_if and replaced all usage.  I also got the vec! macro from alloc::vec and reverted all the macro removal changes, so the diff is much smaller.

Let me know if you need anything else.

xoloki:
Yeah @burdges I really don't know what the point of cfg_if is either, but the first no_std compliant code I looked at used it so I've always done so.  Maybe recent rust modified the #cfg directive so cfg_if is no longer necessary?  Either way, I removed it.

hdevalence:
Great! I tested depending on the `bulletproofs` crate from a scratch library, and building that library in `no_std` mode (to prevent criterion or dev-dependencies from re-enabling std), and ran into a few hitches:

* The `clear_on_drop` crate should probably be selected as 
```
clear_on_drop = { version = "0.2", default-features = false, features = ["nightly"] }
```

* The `rand` crate should be an optional dependency, enabled by the `std` feature of Bulletproofs:
```
rand = { version = "0.6", default-features = false, optional = true }
...
std = ["rand", "rand/std", ...
```

* The crate has an `alloc` feature that enables the imports of all of the alloc-related functionality.  But the crate can't work without that functionality, so compiling with `--no-default-features` fails.  Since we're not going to have an alloc-free implementation, we know we'll always need alloc, so I think it would be better to remove the `alloc` feature and use alloc unconditionally, so that `--no-default-features` selects `no_std + alloc`, instead of breaking the build.

xoloki:
Ok @hdevalence I fixed the clear_on_drop and rand dependencies as you wanted.

However, I'm not sure how to have --no-default-dependencies always select no_std + alloc, can you point me to something that does this?

xoloki:
I could do something like this @hdevalence 

```
#![cfg_attr(not(feature = "std"), feature(alloc))]
```

but wouldn't I still need to define a Cargo.toml alloc feature so I could chain it down to rand_core, serde, and curve25519?

xoloki:
BTW the CI failures appear to be spurious, it was just failure to connect to github:
```
fatal: unable to access 'https://github.com/dalek-cryptography/bulletproofs.git/': Failed to connect to github.com port 443: Connection timed out
The command "eval git clone --depth=50 https://github.com/dalek-cryptography/bulletproofs.git dalek-cryptography/bulletproofs " failed. Retrying, 2 of 3.
```

xoloki:
Okay @hdevalence I got rid of the alloc feature and just used alloc everywhere.  Building without default features just works now.

Let me know if you want me to do anything else.

xoloki:
All tests passing now @hdevalence

src/range_proof/dealer.rs
```diff
@@ -295,15 +302,19 @@ impl<'a, 'b> DealerAwaitingProofShares<'a, 'b> {
     /// performing local aggregation,
     /// [`receive_trusted_shares`](DealerAwaitingProofShares::receive_trusted_shares)
     /// saves time by skipping verification of the aggregated proof.
-    pub fn receive_shares(mut self, proof_shares: &[ProofShare]) -> Result<RangeProof, MPCError> {
+    pub fn receive_shares<T: RngCore + CryptoRng>(
```
hdevalence: I guess this needs to be a new `receive_shares_with_rng` function, because `receive_shares` is part of the public API and can't change without a major version bump.

src/range_proof/party.rs
```diff
@@ -155,12 +155,11 @@ pub struct PartyAwaitingBitChallenge<'a> {
 impl<'a> PartyAwaitingBitChallenge<'a> {
     /// Receive a [`BitChallenge`] from the dealer and use it to
     /// compute commitments to the party's polynomial coefficients.
-    pub fn apply_challenge(
+    pub fn apply_challenge<T: RngCore + CryptoRng>(
```
hdevalence: `apply_challenge` is also stable

src/range_proof/dealer.rs
xoloki: Ok, I'll rework these so the API doesn't change in std mode.

src/range_proof/party.rs
```diff
@@ -68,20 +70,18 @@ pub struct PartyAwaitingPosition<'a> {
 impl<'a> PartyAwaitingPosition<'a> {
     /// Assigns a position in the aggregated proof to this party,
     /// allowing the party to commit to the bits of their value.
-    pub fn assign_position(
+    pub fn assign_position<T: RngCore + CryptoRng>(
```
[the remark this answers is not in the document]
xoloki: Will do.

src/range_proof/party.rs
xoloki: Ok, I'll make these changes today.

hdevalence:
Hey @xoloki, this looks great, I would like to merge it, thanks for all this work!  

I was doing a final pass over the combined changes and I spotted something I should have noticed before, but didn't: the functions in the `aggregation` module that were changed to take RNG parameters are part of the public API of the crate, so they can't be changed without breaking stability.  I'm sorry I didn't spot it before and that there's another round of back-and-forth, but I think that that is the only thing left.

xoloki:
No worries, thanks for all your help getting this in shape.

src/range_proof/dealer.rs
hdevalence: Thanks! Sorry again for missing this on the first pass of review :disappointed:

xoloki:
Ok @hdevalence I fixed the public API to be backwards compatible in std mode, and renamed the nostd public API elements _with_rng.

Let me know if you need any more changes.

src/util.rs
```diff
@@ -64,7 +68,7 @@ pub fn exp_iter(x: Scalar) -> ScalarExp {
 pub fn add_vec(a: &[Scalar], b: &[Scalar]) -> Vec<Scalar> {
     if a.len() != b.len() {
         // throw some error
-        println!("lengths of vectors don't match for vector addition");
+        //println!("lengths of vectors don't match for vector addition");
```
oleganza: Maybe add panic?

hdevalence: I think we can probably clean this up in a different PR

src/range_proof/party.rs
```diff
@@ -155,12 +168,21 @@ pub struct PartyAwaitingBitChallenge<'a> {
 impl<'a> PartyAwaitingBitChallenge<'a> {
     /// Receive a [`BitChallenge`] from the dealer and use it to
     /// compute commitments to the party's polynomial coefficients.
-    pub fn apply_challenge(
+    #[cfg(feature = "std")]
+    pub fn apply_challenge<T: RngCore + CryptoRng>(
```
hdevalence:
```suggestion
    pub fn apply_challenge(
```

xoloki: Done, good catch!

src/range_proof/mod.rs
```diff
@@ -613,7 +714,9 @@ mod tests {
 
     #[test]
     fn create_and_verify_n_64_m_8() {
-        singleparty_create_and_verify_helper(64, 8);
+        // XXX disable this test since it's failing with
+        //     'attempt to negate with overflow'
+        //singleparty_create_and_verify_helper(64, 8);
```
hdevalence: This seems kind of worrying (in the sense that it may point at a problem in the rest of the source code) -- is it reproducible? I tried to reproduce on my machine but I wasn't able to.

hdevalence:
I have a cherry-pick'able commit on a copy of this branch (https://github.com/hdevalence/bulletproofs/commit/41917811c2eb856c5846c75ebdcc39e78488d687 ) that undoes the changes to the test code, so that the test code continues to use the previous aggregation API. This way we test that the previous API still works correctly.

src/range_proof/mod.rs
xoloki: Weird, I thought I reverted this.  It was failing locally, but not remotely in CI.  

Reverted now.

xoloki:
> I have a cherry-pick'able commit on a copy of this branch ([hdevalence@4191781](https://github.com/hdevalence/bulletproofs/commit/41917811c2eb856c5846c75ebdcc39e78488d687) ) that undoes the changes to the test code, so that the test code continues to use the previous aggregation API. This way we test that the previous API still works correctly.

The cherry has been picked, thanks!

hdevalence:
Hey @xoloki -- I was on vacation so I missed this; I'll try to get this merged before the end of the week!

xoloki:
Excellent, thanks @hdevalence!

hdevalence:
Planning to get this merged this afternoon!

hdevalence:
Hey @xoloki, I think I'm out of time to do a release for this today but I will do it at the beginning of next week.  Sorry that this took so long and thanks so much for your work on this :)

xoloki:
Thanks for the merge!  Next week is totally fine for a release.

PR #284, opened 2019-07-29
Add ConstraintSystem:: multipliers_len()

rubdos:
For some benchmarks and tests it is useful to retrieve the amount of
actually allocated multipliers from the proof system.

This is especially useful when building complex and composed circuits, when one doesn't want to count them manually :-)


oleganza:
Oh, sorry for having this sitting for so long. Just one nitpick: maybe rename to `multipliers_count`, so it does not sound like it starts a slow and complicated counting procedure?

rubdos:
> Oh, sorry for having this sitting for so long. Just one nitpick: maybe rename to `multipliers_count`, so it does not sound like it starts a slow and complicated counting procedure?

I've seen another repository where they used `num_multipliers`, and I was just thinking about `multipliers_len()` or `len_multipliers()`, in accordance with `Vec::len()`. what would you prefer?

oleganza:
`multipliers_len` also works!

Issue #283, opened 2019-07-29
Incorrect path for doc include range proof


vlopes11:
Seems I'm having some problems to evaluate the path for the doc includes with the latest Rust.

```
error: couldn't read src/range_proof/../docs/range-proof-protocol.md: No such file or directory (os error 2)
 --> src/range_proof/mod.rs:2:18
  |
2 | #![doc(include = "../docs/range-proof-protocol.md")]
  |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ couldn't read file
```

The absolute path should consider one more level. Here is a pull request that fixes it.
https://github.com/dalek-cryptography/bulletproofs/pull/282

hdevalence:
Thanks for spotting this -- I think that there might have been a change to doc import paths.  Right now we're in a bit of a weird state because the travis config is using an outdated nightly, so I'll circle back to this after updating it.

hdevalence:
https://rust-lang.github.io/rustup-components-history/ suggests that all of the components we like (rls, fmt, etc) are available as of today's nightly, let's try upgrading.

vlopes11:
Tested and fixed. I'll close the PR + this issue.

Thanks!

PR #290, opened 2019-07-31
updated the path to fix compilation error

ssahai:
Fixing the following error:

error: couldn't read src/range_proof/../docs/range-proof-protocol.md: No such file or directory (os error 2)
 --> src/range_proof/mod.rs:2:18
  |
2 | #![doc(include = "../docs/range-proof-protocol.md")]
  |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ couldn't read file

### DIFF
--- a/src/range_proof/mod.rs
+++ b/src/range_proof/mod.rs
@@ -1,5 +1,5 @@
 #![allow(non_snake_case)]
-#![doc(include = "../docs/range-proof-protocol.md")]
+#![doc(include = "../../docs/range-proof-protocol.md")]
 
 use rand;
 


hdevalence:
Hey, thanks for sending this PR!

I'm going to close it, though, because it's a duplicate of #283 -- the only reason that one hasn't been merged yet is that this change to import path roots for doc(include) happened recently, and we need to check which nightly version so that we can update the CI pin (#287) so the CI isn't out of sync, and that required fixing up #277 because the development branch got out of sync.

Thanks for your understanding :)

PR #296, opened 2019-08-10
Better validation of proof shares in Range Proof MPC

oleganza:
Addressing issues raised by the quarkslab audit (to be released soon):

* additional size validation for individual proof shares,
* correct `size_hint` for `AggregatedGensIter`,
* reused computation of powers-of-z in the proof aggregation to save a bit of CPU.

### DIFF
--- a/src/range_proof/dealer.rs
+++ b/src/range_proof/dealer.rs
@@ -218,6 +218,15 @@ impl<'a, 'b> DealerAwaitingProofShares<'a, 'b> {
             return Err(MPCError::WrongNumProofShares);
         }
 
+        // Validate lengths for each share
+        for (j,share) in proof_shares.iter().enumerate() {
+            let n = share.check_size(&self.bp_gens, j).map_err(|_| MPCError::WrongNumProofShares)?;
+            if n != self.n {
+                // TBD: better error!
+                return Err(MPCError::WrongNumProofShares);
+            }
+        }
+
         let t_x: Scalar = proof_shares.iter().map(|ps| ps.t_x).sum();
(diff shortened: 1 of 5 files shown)


cathieyun:
👀

cathieyun:
I added the requirement to use minimum 1.2.3 for curve25519-dalek, while we're doing safety tweaks, per @hdevalence 's suggestion.

The share validation updates look good!

oleganza:
Thanks @cathieyun!

src/range_proof/dealer.rs
```diff
@@ -218,6 +218,27 @@ impl<'a, 'b> DealerAwaitingProofShares<'a, 'b> {
             return Err(MPCError::WrongNumProofShares);
         }
 
+        // Validate lengths for each share
+        let mut bad_shares = Vec::<usize>::new(); // no allocations until we append
+        for (j, share) in proof_shares.iter().enumerate() {
+            match share.check_size(&self.bp_gens, j).and_then(|n| {
```
hdevalence: Instead of having `check_size` return a `Result` with the actual `n` parameter, I think it would be better to have `check_size` take an `expected_n: usize` and fold the actual n == expected n check into the rest of the size check function, so that all of the consistency checking logic is in one place. This also makes the check_size calling code simpler (it might even be possible to do something like `.map_err(|_| bad_shares.push(j))` but I'm not sure if the borrow checker will be happy about a &mut closure capture).

oleganza: Good point, will do.

oleganza: image

PR #298, opened 2019-08-21
Strip down allocation in serialization of R1CSProof

AnthonyMikh:
This PR adds `InnerProductProof::to_bytes_iter` which similar to `proof.to_bytes().into_iter()`, but doesn't allocate, and uses it in serialization of `R1CSProof`.

Unresolved questions:

- [ ] Should it be public API?

### DIFF
--- a/src/inner_product_proof.rs
+++ b/src/inner_product_proof.rs
@@ -346,6 +346,20 @@ impl InnerProductProof {
         buf
     }
 
+    /// Converts the proof into a byte iterator over serialized view of the proof.
+    /// The layout of the inner product proof is:
+    /// * \\(n\\) pairs of compressed Ristretto points \\(L_0, R_0 \dots, L_{n-1}, R_{n-1}\\),
+    /// * two scalars \\(a, b\\).
+    #[inline]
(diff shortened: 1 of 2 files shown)


oleganza:
LGTM, i'd only make it `pub(crate)` for now. In a separate PR we can roll out public `to_bytes_iter()` APIs for both IPP, R1CS and rangeproofs. Then we'll probably know how to better name all of them.

src/range_proof/mod.rs
```diff
@@ -393,8 +393,7 @@ impl RangeProof {
         buf.extend_from_slice(self.t_x.as_bytes());
         buf.extend_from_slice(self.t_x_blinding.as_bytes());
         buf.extend_from_slice(self.e_blinding.as_bytes());
-        // XXX this costs an extra alloc
-        buf.extend_from_slice(self.ipp_proof.to_bytes().as_slice());
+        buf.extend_from_slice(self.ipp_proof.to_bytes_iter());
```
oleganza: should this be `.extend`?

```suggestion
        buf.extend(self.ipp_proof.to_bytes_iter());
```

PR #303, opened 2019-11-26
Update rand_core, curve25519-dalek, merlin versions.

hdevalence:
This is a breaking API change that will require a `2.0.0` version, but it keeps the crate up-to-date with the rand ecosystem. A possible further simplification of the dep tree would be to replace `clear_on_drop` by `zeroize`, but this could be done at any later point because `clear_on_drop` is an implementation detail.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,24 +12,24 @@ keywords = ["cryptography", "crypto", "ristretto", "zero-knowledge", "bulletproo
 description = "A pure-Rust implementation of Bulletproofs using Ristretto"
 
 [dependencies]
-curve25519-dalek = { version = "^1.2.3", default-features = false, features = ["u64_backend", "nightly", "serde", "alloc"] }
+curve25519-dalek = { version = "2", default-features = false, features = ["u64_backend", "nightly", "serde", "alloc"] }
 subtle = { version = "2", default-features = false }
 sha3 = { version = "0.8", default-features = false }
 digest = { version = "0.8", default-features = false }
-rand_core = { version = "0.4", default-features = false, features = ["alloc"] }
-rand = { version = "0.6", default-features = false, optional = true }
+rand_core = { version = "0.5", default-features = false, features = ["alloc"] }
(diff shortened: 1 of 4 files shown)


oleganza:
I've also bumped criterion dependency to 0.3 - seems like it's compatible with our benchmarks. So it uses the same `rand` as the rest of the code. Otherwise we'd have 3 different versions of `rand` in benchmark builds.

hdevalence:
Yay!

hdevalence:
I guess the only thing to remember is that when we do a 2.0 release we should be sure to disable the `yoloproofs` feature again in the published crate.

oleganza:
>I guess the only thing to remember is that when we do a 2.0 release we should be sure to disable the yoloproofs feature again in the published crate.

is it done manually somehow?

hdevalence:
Yeah, last time the workflow was (assuming `main`, `develop` are in sync with the repo):

1. Branch `develop` to `release/X.Y.Z`
2. In the `release` branch, comment out `yoloproofs` in the `Cargo.toml` in a single atomic commit
3. Add other commits updating the changelog, version number, etc.
4. Tag the release as `X.Y.Z`, checkout the tag, `cargo publish`
5. Merge `release/X.Y.Z` into `main`
6. Merge `main` into `develop`
7. On `develop`, do a revert of the atomic commit from (2).
8. Push `main`, `develop`, and the `X.Y.Z` tag to the repo.

oleganza:
_if CI is so damn clever to check rustfmt, can't just do it all by itself, check that tests work and offer you a one-click green-checkmark commit button without waiting for the tests to run all over again_

oleganza:
maybe in a couple of months we'll be able to get CI finish in green

hdevalence:
I tried manually restarting the build.

hdevalence:
The build timed out again, but running the command locally passes. My preference would be to merge this and try to fix the CI later, does that sound good?

oleganza:
Yup!

hdevalence:
Done!

PR #305, opened 2019-12-12
Curve v2 (on develop branch)

cbeck88:
Uprev curve25519-dalek to v2 and use matching rand and rand_core versions (matching current master)

We need this because we are using bulletproofs/develop branch (for no_std support) and we also need to uprev curve25519-dalek to v2 (because the old version has a build.rs and depends on byteorder, but we also have build-dependencies that turn on byteorder/std, breaking our no_std build)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,12 +12,12 @@ keywords = ["cryptography", "crypto", "ristretto", "zero-knowledge", "bulletproo
 description = "A pure-Rust implementation of Bulletproofs using Ristretto"
 
 [dependencies]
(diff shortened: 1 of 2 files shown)


cbeck88:
it appears that this breaks yoloproofs unless we will also uprev rand_core used in merlin crate

hdevalence:
Hi, it looks like this might be a duplicate of https://github.com/dalek-cryptography/bulletproofs/pull/303 ?

oleganza:
@cbeck88 thanks for bumping this issue. I went to fix #303, so we hopefully will fully migrate to v2 curve-dalek after CI finishes.

hdevalence:
These changes are included in #303, thanks!

PR #308, opened 2019-12-27
Replace failure with thiserror to have std-compatible errors

### LINKED ISSUE #306: Remove `failure` and use `std::error::Error` via `thiserror`
Since we need a 2.0.0 release due to #303, we could also use this opportunity to migrate away from `failure` (now that `std::error::Error` has been improved).

hdevalence:
Because std::error::Error is not part of core, this commit feature-gates the
derived `Error` implementation on whether `std` is available.  If we are in a
`no_std` context, the `Error` enums are just plain enums with no `Error`
implementation.

Closes #306 

Follows on from #307, so this should be merged only after that one is.


hdevalence:
Rebased onto current `develop`.

hdevalence:
Fixed a missing `display` and feature-gating on the serde implementations in the R1CS module.

PR #309, opened 2020-01-08
Fix latex render

contrun:
There are current some render problem on page https://doc-internal.dalek.rs/bulletproofs/notes/r1cs_proof/index.html.
screenshot-2020-01-08-15-49-00

This pr should fix that.
screenshot-2020-01-08-15-49-50


contrun:
Oops. Just made a pr to the main branch. Was intended to be develop.

PR #313, opened 2020-03-12
r1cs: accept ownable transcript

oleganza:
This changes the Prover/Verifier API as follows:

1. Instead of holding `&mut Transcript`, Prover/Verifier accepts `BorrowMut<Transcript>`. This means, Transcript can be passed in not only as a mutable reference, but also moved in and owned by the constraint system.
2. `Verifier::verify` and `Prover::prove` have new variants `_and_return_transcript()` that return the transcript back to the user. So if the user moved the transcript into CS, they can get it back and continue working with it.

### Rationale

The proposed API allows users to "bundle" transcript and intermediate CS state in an object, so that it can be used after the CS is built, but before the proof is created/verified. Specifically, in [ZkVM](https://github.com/stellar/slingshot/blob/main/zkvm/src/verifier.rs), the transaction effects and its ID are computed together with building a constraint system. New API would allow us to perform cheaper verification of the effects before doing a more expensive ZKP verification. Without the new API we'd need to make an extra precomputation pass to get the effects w/o ZKP verification.

### Alternative: pinned bundle around the existing API

We can probably keep the self-referential struct: the bundle would have to own a transcript and also contain a CS instance with a &mut reference to it. To make such object safe to use, it must be additionally wrapped in a pinned Box, and also we'd need some careful pointer manipulation inside `unsafe{}`. I haven't tried it out, but it does not seem very straightforward.

Pros: the API remains the same.
Cons: more complex wrapper solution.

### DIFF
--- a/src/r1cs/prover.rs
+++ b/src/r1cs/prover.rs
@@ -1,6 +1,7 @@
 #![allow(non_snake_case)]
 
 use clear_on_drop::clear::Clear;
+use core::borrow::BorrowMut;
 use core::mem;
 use curve25519_dalek::ristretto::{CompressedRistretto, RistrettoPoint};
 use curve25519_dalek::scalar::Scalar;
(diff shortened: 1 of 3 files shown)


src/r1cs/prover.rs
```diff
@@ -101,13 +109,13 @@ impl<'t, 'g> ConstraintSystem for Prover<'t, 'g> {
         let o = l * r;
```
WaffleLapkin: Here (and in some other places) It's possible to extract fields of `Secrets` to remove `self.secrets.<...>` boilerplate.
```rust
let Secrets { a_L, a_R, a_O, .. } = &mut self.secrets
```
Mb this will make the code a bit cleaner

PR #318, opened 2020-04-26
Support for stable Rust

oleganza:
This adds support for using stable Rust by introducing a new cargo feature `nightly`. Since all the docs are using unstable `external_doc` compiler feature, the docs can only be build on a nightly Rust.

```
# basic features on stable:
cargo build

# with R1CS API on stable:
cargo build --features="yoloproofs"

# no_std (nightly only):
cargo build --no-default-features --features="nightly"

# vectorized backend (nightly only):
cargo build --no-default-features --features="avx2_backend"

# build documentation (nightly only)
make doc-internal
```

Since Travis CI is not reliable anymore, we also move to Github Actions. Both `main` and `develop` branches are checked against `stable` and `nightly` Rust. Main branch does not use `yoloproofs` feature.

For development, we keep `rust-toolchain` with the latest available nightly. If it ever breaks, one can switch over to stable instead of hunting down a more workable version of nightly (although, docs won't build).

(Replaces #314)


hdevalence:
Didn't test, but this looks great to me.

oleganza:
image

i think that's the end for Travis

PR #322, opened 2020-05-14
R1CS: Add `constraints_len`, `deferred_constraints_len` and `total_constraints_len`

rubdos:
Add methods to count the different kinds of constraints, which comes in handy when developing and benchmarking larger circuits. This is analogue with #284, `ConstraintSystem::multipliers_len()`.

### DIFF
--- a/src/r1cs/constraint_system.rs
+++ b/src/r1cs/constraint_system.rs
@@ -66,8 +66,8 @@ pub trait ConstraintSystem {
         input_assignments: Option<(Scalar, Scalar)>,
     ) -> Result<(Variable, Variable, Variable), R1CSError>;
 
-    /// Counts the amount of allocated multipliers.
-    fn multipliers_len(&self) -> usize;
+    /// Counts the amount of constraints in the constraint system.
+    fn metrics(&self) -> crate::r1cs::Metrics;
 
     /// Enforce the explicit constraint that
     /// ```text
--- a/src/r1cs/metrics.rs
+++ b/src/r1cs/metrics.rs
@@ -1,0 +1,10 @@
+/// A struct that contains metrics about a constraint system.
+///
+/// See [`ConstraintSystem::metrics`](::r1cs::ConstraintSystem::metrics).
+#[derive(Debug, Clone)]
+pub struct Metrics {
+    pub multipliers: usize,
+    pub constraints: usize,
+    pub phase_one_constraints: usize,
+    pub phase_two_constraints: usize,
+}
--- a/src/r1cs/metrics.rs
+++ b/src/r1cs/metrics.rs
@@ -3,8 +3,12 @@
 /// See [`ConstraintSystem::metrics`](::r1cs::ConstraintSystem::metrics).
 #[derive(Debug, Clone)]
 pub struct Metrics {
+    /// Number of multiplicative constraints in the constraint system.
     pub multipliers: usize,
+    /// Total number of linear constraints in the constraint system.
     pub constraints: usize,
+    /// Number of linear constraints added in pre-randomization phase.
     pub phase_one_constraints: usize,
+    /// Number of linear constraints added in the randomization phase.
     pub phase_two_constraints: usize,
 }
--- a/src/r1cs/mod.rs
+++ b/src/r1cs/mod.rs
@@ -5,6 +5,7 @@ mod notes {}
 
 mod constraint_system;
 mod linear_combination;
+mod metrics;
 mod proof;
 mod prover;
 mod verifier;
@@ -13,6 +14,7 @@ pub use self::constraint_system::{
     ConstraintSystem, RandomizableConstraintSystem, RandomizedConstraintSystem,
 };
 pub use self::linear_combination::{LinearCombination, Variable};
+pub use self::metrics::Metrics;
 pub use self::proof::R1CSProof;
 pub use self::prover::Prover;
 pub use self::verifier::Verifier;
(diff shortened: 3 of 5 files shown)


src/r1cs/constraint_system.rs
```diff
@@ -69,6 +69,17 @@ pub trait ConstraintSystem {
     /// Counts the amount of allocated multipliers.
     fn multipliers_len(&self) -> usize;
 
+    /// Counts the amount of first-phase constraints in the circuit.
+    fn constraints_len(&self) -> usize;
```
oleganza, line 73: To my taste, i'd leave this generic sounding name for all the constraints, not just the first phase. If it's important to distinguish between first phase/second phase, I'd have specific methods for them.

Also: it may be useful to just have one method `stats() -> Stats` that returns a non-exhaustive struct (with private PhantomData or smth like that?) that contains `.multipliers: usize, .constraints: usize, first_phase_constraints: usize, .second_phase_constraints: usize` etc.

rubdos: Ack on the naming, I'll swap that out.

The most valuable thing about such a `Stats` struct would be its (derived) `Debug` implementation. On the other hand, these methods are not really to be "used in production" anyway, so having a separate struct may be a tid of overengineering...

Anyhow, your pick, let me know what you prefer!

oleganza: For debugging it seems to be more useful to have all stats/metrics in one nifty struct, so we don't have to fiddle with calling individual methods. We don't need obnoxious API for that thing, other than pub fields and derive(Debug).

Having a private field is only to be future-proof and be able to add more fields w/o breaking people's code.

oleganza: Feel free to move the current stats under that struct and break the API.

rubdos: Maybe another question: what would be the preferred method and struct name? `Stats` sounds like it's a statistic, while everything is a constant. Should I prefer `Metrics` or even `CircuitMetrics` or something like that?

Beyond multipliers and constraint counts, I don't think there's other metrics to include, right?

Another question: the private PhantomData you refer to, why would the `Stats` struct need to be generic; do you mean should it reference to the ConstraintSystem, such that an instance of `Stats` is always correct (by making the CS immutable for the lifetime of the Stats instance)?

oleganza: Yeah, something alongside "metrics" sounds better. I'd prefer to avoid using the term "circuit" since we are not using it anywhere in the API. Maybe simply `r1cs::Metrics` is good enough?

As of PhantomData: sorry, i meant having a dummy private field with a zero-sized type, so the struct cannot be used in any other way than just reading its fields. E.g. no destructuring etc.

```
#[derive(Debug)]
struct Metrics {
    pub multipliers_count: usize,
    pub constraints_count: usize,
    _private: ()
}
```

But it makes Debug printout uglier and you cannot opt out easily. Maybe we should just list the fields w/o any extra precautions.

rubdos:
Made a `r1cs::Metrics` struct instead!

src/r1cs/constraint_system.rs
rubdos: FWIW, I'd leave destructuring open for the user. It may be handy for writing benchmarks, etc; I don't think it leaves a lot open for abuse.

rubdos:
Squashed together; didn't make any sense anymore as separate commits. Note that this removes `multipliers_len`, as suggested by @oleganza .

src/r1cs/metrics.rs
```diff
@@ -0,0 +1,10 @@
+/// A struct that contains metrics about a constraint system.
+///
+/// See [`ConstraintSystem::metrics`](::r1cs::ConstraintSystem::metrics).
+#[derive(Debug, Clone)]
+pub struct Metrics {
+    pub multipliers: usize,
```
oleganza, line 6:
```suggestion
    /// Number of multiplicative constraints in the constraint system
    pub multipliers: usize,
```

src/r1cs/metrics.rs
```diff
@@ -0,0 +1,10 @@
...
+    pub constraints: usize,
```
oleganza, line 7:
```suggestion
    /// Total number of linear constraints in the constraint system
    pub constraints: usize,
```

src/r1cs/metrics.rs
```diff
@@ -0,0 +1,10 @@
...
+    pub phase_one_constraints: usize,
```
oleganza, line 8:
```suggestion
    /// Number of linear constraints added in pre-randomization phase.
    pub phase_one_constraints: usize,
```

src/r1cs/metrics.rs
```diff
@@ -0,0 +1,10 @@
...
+    pub phase_two_constraints: usize,
```
oleganza, line 9:
```suggestion
    /// Number of linear constraints added in the randomization phase.
    pub phase_two_constraints: usize,
```

oleganza:
I've added documentation suggestions - please feel free to reword how you please. This patch looks great!

rubdos:
I hadn't seen your update, I have accepted your suggestions!

oleganza:
Nightly is already broken for everyone using dalek libraries since they renamed `asm!` macro to `llvm_asm!`, which breaks `subtle` (https://github.com/dalek-cryptography/subtle/pull/70), so I'm merging this PR even with CI failing on nightly.

PR #325, opened 2020-06-10
Use a more verbose assert!() in r1cs docs example.

rubdos:
While debugging complex circuits, the actual R1CSError value does not become apparent while debugging. It might be difficult for newcomers to Bulletproofs to distinguish between the error condition of having written a faulty gadget (`proof.verify(...)` propagates `R1CSError`s through the `?` operator), versus not having allocated enough multipliers.

I am talking from experience here: I was looking at code, thinking my own gadget was failing, while in fact the assertion was copy-pasted from the example code in the docs!  Changing this, it became apparent that we needed to extend BulletproofGens a bit.


rubdos:
I am unsure whether that fragment gets doc-tested, maybe it needs a second look.

Issue #278, opened 2019-06-06
ConstraintSystem::constrain(): check validity of LinearCombinations.


rubdos:
I found these lines in `prover.rs`:

```rust
fn constrain(&mut self, lc: LinearCombination) {
// TODO: check that the linear combinations are valid
// (e.g. that variables are valid, that the linear combination evals to 0 for prover, etc).
//[..]
}
```

I tried implementing this, since I have some issues debugging a proof system I built. I've written:

```rust
    fn constrain(&mut self, lc: LinearCombination) {
        let sum: Scalar = lc.terms.iter().map(|(v, s)| {
            let l = match v {
                Variable::One() => Scalar::one(),
                Variable::Committed(i) => self.v[*i],
                Variable::MultiplierLeft(i) => self.a_L[*i],
                Variable::MultiplierRight(i) => self.a_R[*i],
                Variable::MultiplierOutput(i) => self.a_O[*i],
            };
            l * s
        }).sum();
        debug_assert_eq!(sum, Scalar::zero());
        // TODO: check that the linear combinations are valid
        // (e.g. that variables are valid, that the linear combination evals to 0 for prover, etc).
        self.constraints.push(lc);
    }
}
```

Which *works* but it changes the behaviour of the method. Certain tests (e.g. `example_gadget_test()`) started failing, because now the prove system does not return `Err` on validation, but panics earlier.

---

Another option would be to return `Result<(), R1CSError>` in `fn constrain`, but that alters the signature for `ConstraintSystem` (and thus for verifier, who will always return `Ok(())` anyhow). It would also not allow to disable this behaviour on `--release` builds.

---

What kind of implementation for this `TODO` would you consider best? Maybe even something different from my proposals?

PR #326, opened 2020-06-11
Set `Prover::eval` to pub, for easier debugging of circuits.

rubdos:
Development of complex constraint systems often goes wrong, and this patch allows to evaluate (for the prover) the actual value of a LinearCombination.

Related to #278, if it doesn't even fix that.

### DIFF
--- a/src/r1cs/prover.rs
+++ b/src/r1cs/prover.rs
@@ -354,6 +354,7 @@ impl<'g, T: BorrowMut<Transcript>> Prover<'g, T> {
         (wL, wR, wO, wV)
     }
 
+    /// Returns the secret value of the linear combination.
     pub fn eval(&self, lc: &LinearCombination) -> Scalar {
         lc.terms
             .iter()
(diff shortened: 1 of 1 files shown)


rubdos:
Just noticed there's `eval` already... I'll change this mr to make that pub.

src/r1cs/prover.rs
```diff
@@ -354,7 +354,7 @@ impl<'g, T: BorrowMut<Transcript>> Prover<'g, T> {
         (wL, wR, wO, wV)
     }
 
-    fn eval(&self, lc: &LinearCombination) -> Scalar {
+    pub fn eval(&self, lc: &LinearCombination) -> Scalar {
```
oleganza, line 357:
```suggestion
    /// Returns the secret value of the linear combination.
    pub fn eval(&self, lc: &LinearCombination) -> Scalar {
```

rubdos:
I should've noticed that I had to add a doc comment... Either way, does this need someone else to review or would this be good to go? :-)

oleganza:
Yeah, i was waiting for CI and then went to sleep.

PR #328, opened 2020-06-25
Update rust-toolchain to fix check error

Ancient123:
dependency clear_on_drop-0.2.4 expects rust macro `llvm_asm` which the previous nightly did not have. 
Update to 2020-05-15 which is the latest version that supports *both* llvm_asm and asm macros, to cause the fewest issues with people.


Ancient123:
@hdevalence Are you planning on merging this?

hdevalence:
Oops, thanks for the reminder! Will do this today.

PR #330, opened 2020-07-15
Serde optional and `Clear_on_drop` replaced by `Zeroize`

fmiguelgarcia:
In order to improve the support of  `no_std` and another platforms like WASM:  
* `Serde` is optional
* Replace `clear_on_drop` by `zeroize`:  No inline LLVM asm and WASM support..
* Support different back-ends: `u32_backend`, `u64_backend`, and `avx2_backend`(default in `std`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 Entries are listed in reverse chronological order.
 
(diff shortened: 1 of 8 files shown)


gedigi:
can this be merged?

Issue #335, opened 2020-10-21
Updated: Bulletproof rewinding proposed for a single range proof


hansieodendaal:
This is a proposal to add Bulletproofs rewinding functionality to the Bulletproofs crate as a user option. (updates #329)

We use `dalek-cryptography/bulletproofs` in our Mimblewimble implementation project and have a need to do wallet recovery from seed values. The proposal for the Bulletproof rewinding scheme is implemented [here, in a fork on main](https://github.com/hansieodendaal/bulletproofs/tree/investigate_rewind), and demonstrated from a user perspective [in this test](https://github.com/hansieodendaal/bulletproofs/blob/investigate_rewind/tests/range_proof.rs#L154).
_**Note:** This proposal is similar to what has been proposed and implemented for Grin, [as discussed here](https://github.com/mimblewimble/grin-wallet/issues/105)._

Three additional parameters are introduced when creating a proof for a value commitment:
- `pvt_rewind_key: &Scalar`;
- `pvt_blinding_key: &Scalar` and 
- `proof_message: &[u8; 23]`.

The 23 bytes worth of proof message can be any message a user wants to embed within the proof. Internally the two private keys, in combination with the value commitment, are converted into two rewind nonces and two blinding nonces:
``` rust
rewind_nonce_1 = H( H(pub_key_from_pvt_key(pvt_rewind_key)), value_commitment)
rewind_nonce_2 = H( H(pub_key_from_pvt_key(pvt_blinding_key)), value_commitment)
blinding_nonce_1 = H( H(pvt_rewind_key), value_commitment)
blinding_nonce_2 = H( H(pvt_blinding_key), value_commitment)
```

With the [Party and Dealer's algorithm](https://doc-internal.dalek.rs/bulletproofs/range_proof/index.html#party-and-dealers-algorithm):

![image](https://user-images.githubusercontent.com/39146854/86337922-070f3a00-bc52-11ea-9067-bff87e0db13a.png)

1. `a_blinding` is replaced by `rewind_nonce_1`
1. `s_blinding` is replaced by `XOR(rewind_nonce_2, merge_into_word(value, proof_message))`

![image](https://user-images.githubusercontent.com/39146854/86339020-5dc94380-bc53-11ea-8948-bf97535f5641.png)

1. `t_1_blinding` is replaced by `blinding_nonce_1`
1. `t_2_blinding` is replaced by `blinding_nonce_2`

Usage:
[...]

PR #340, opened 2021-01-08
Add Bulletproof rewind functionality

hansieodendaal:
This is closely modelled on Grin's solution, but using two private keys:
- Interfaces:
  - create a rewindable ZK proof with up to 23 bytes additional embedded proof
    data
  - extract the value and 23 bytes proof data only
  - extract the value, blinding factor and 23 bytes proof data
- Required:
  - two rewind nonces, based on two public rewind keys and the value
    commitment
  - two blinding nonces, based on the two private rewind keys and the value
    commitment
- Use:
  - the two rewind nonces are used to extract the value and 23 bytes embedded
    proof data
  - the two rewind nonces and two blinding nonces are used to extract the
    value, blinding factor and 23 bytes embedded proof data

Please see related issue #335 

Thanks!

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -45,6 +45,18 @@ pub enum ProofError {
     /// consider its errors to be internal errors.
     #[cfg_attr(feature = "std", error("Internal error during proof creation: {0}"))]
     ProvingError(MPCError),
+    /// This error results from trying to rewind a proof with the wrong rewind nonce
+    #[cfg_attr(
+    feature = "std",
+    error("Rewinding the proof failed, invalid commitment extracted")
+    )]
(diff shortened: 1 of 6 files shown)


delta1:
Hi bulletproofs team! @hdevalence @cathieyun @isislovecruft @oleganza 

Could we have any feedback on this proposal/PR please? We are maintaining a fork over at [tari-project/bulletproofs](/tari-project/bulletproofs), but would love to have this merged upstream if it's something you find could be useful. 

Any comment or critique is greatly welcomed. Thank you

delta1:
Hello @hdevalence @cathieyun @isislovecruft @oleganza do you have any feedback please? Thank you for your work

PR #358, opened 2022-09-07
Upgrade to 2021 edition

rickwebiii:


rillian (commented):
Looks reasonable to me, although I'd address the `to_bytes` dead code warning by either removing the method or adding something to keep it alive, rather than hiding it.

Issue #363, opened 2023-01-09
Error: external_doc feature removed and unknown feature 


ercembu:
When I add the crate as such in Cargo.toml:
```
[package]
name = "my_pack"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
bulletproofs = {version = "2.0.0" }
```

I get the following errors which seem to be due to some removed feature from rust?
```
$ cargo check
    Checking bulletproofs v2.0.0
error[E0557]: feature has been removed
 --> /home/bulut/.cargo/registry/src/github.com-1ecc6299db9ec823/bulletproofs-2.0.0/src/lib.rs:3:12
  |
3 | #![feature(external_doc)]
  |            ^^^^^^^^^^^^ feature has been removed
  |
  = note: use #[doc = include_str!("filename")] instead, which handles macro invocations

error[E0635]: unknown feature `try_trait`
 --> /home/bulut/.cargo/registry/src/github.com-1ecc6299db9ec823/bulletproofs-2.0.0/src/lib.rs:4:12
  |
4 | #![feature(try_trait)]
  |            ^^^^^^^^^

Some errors have detailed explanations: E0557, E0635.
For more information about an error, try `rustc --explain E0557`.
error: could not compile `bulletproofs` due to 2 previous errors

```

The rust is in nightly by default but i don't know if I've missed anything else.

rubdos:
Nightly Rust is usually pinned to a version. rust-toolchain currently lists `nightly-2019-07-31` for this crate.

ercembu:
After moving to the pinned version, now i'm getting errors about the dependencies. 
```
$ cargo check
    Updating crates.io index
    Checking subtle v2.4.1
   Compiling proc-macro2 v1.0.49
    Checking packed_simd_2 v0.3.8
    Checking clear_on_drop v0.2.5
    Checking sha3 v0.8.2
    Checking rand v0.7.3
error[E0658]: use of unstable library feature 'asm': inline assembly is not stable enough for use and is subject to change
  --> /home/bulut/.cargo/registry/src/github.com-1ecc6299db9ec823/clear_on_drop-0.2.5/src/hide.rs:41:17
   |
41 |                 asm!("/* {0} */", in(reg) (ptr as *mut u8), options(nostack));
   |                 ^^^
   |
   = note: for more information, see https://github.com/rust-lang/rust/issues/29722
[...]

Issue #359, opened 2022-10-26
error: could not compile `packed_simd_2` due to 3 previous errors


bo-hub:
# Error report

I git clone this repo and then run cargo test. And the error information came out. 
```
$ cargo test
    Updating crates.io index
  Downloaded serde_derive v1.0.147
  Downloaded serde v1.0.147
  Downloaded serde_json v1.0.87
  Downloaded syn v1.0.103
  Downloaded libc v0.2.137
  Downloaded 5 crates (1.1 MB) in 10.09s
warning: invalid feature `yoloproofs` in required-features of target `r1cs`: `yoloproofs` is not present in [features] section
   Compiling proc-macro2 v1.0.47
   Compiling quote v1.0.21
   Compiling unicode-ident v1.0.5
   Compiling syn v1.0.103
   Compiling autocfg v1.1.0
   Compiling cfg-if v1.0.0
   Compiling libc v0.2.137
   Compiling serde_derive v1.0.147
   Compiling serde v1.0.147
   Compiling typenum v1.15.0
   Compiling crossbeam-utils v0.8.12
   Compiling getrandom v0.1.16
   Compiling memchr v2.5.0
   Compiling unicode-xid v0.2.4
   Compiling libm v0.1.4
   Compiling rayon-core v1.9.3
   Compiling byteorder v1.4.3
   Compiling scopeguard v1.1.0
   Compiling packed_simd_2 v0.3.8
   Compiling serde_json v1.0.87
   Compiling either v1.8.0
   Compiling byte-tools v0.3.1
   Compiling ryu v1.0.11
   Compiling cc v1.0.73
   Compiling unicode-width v0.1.10
   Compiling itoa v1.0.4
   Compiling ppv-lite86 v0.2.16
   Compiling keccak v0.1.2
   Compiling regex-automata v0.1.10
   Compiling plotters-backend v0.3.4
   Compiling thiserror v1.0.37
   Compiling lazy_static v1.4.0
   Compiling same-file v1.0.6
   Compiling subtle v2.4.1
   Compiling bitflags v1.3.2
   Compiling opaque-debug v0.2.3
   Compiling cast v0.3.0
   Compiling itoa v0.4.8
   Compiling half v1.8.2
   Compiling regex-syntax v0.6.27
   Compiling oorandom v11.1.3
   Compiling hex v0.3.2
   Compiling memoffset v0.6.5
   Compiling crossbeam-epoch v0.9.11
[...]

PR #364, opened 2023-01-17
u128 and Scalar

ripa1995:
**Disclaimer:** This pull request has been made against `main` and not `develop` since the latter is unmaintained, while the former has recent updates.

### Motivations

Allowing bigger inputs increases the landscape of domains in which `bulletproofs` can be used.

We increased the `bitsize` to 128 and replaced the `u64` type with `u128` in many places within the code.

We also decided to support `Scalar` as input value, and we decided to implement it as a feature of the library, which can be enabled with `scalar_range_proof`. Also in this case we had to modify the code such that `Scalar` could be accepted. To support `Scalar` we had to implement the extraction of a single bit from a `Scalar`.

Our implementation of the above consists in an indexing of the `Scalar`'s underlying array of bytes, followed by a shift and mask of the appropriate byte. In both cases where this operation is used, the index is not secret, so the timing and memory access patterns of this operation shouldn't enable any side-channel attack.

We think that at the moment there is no bigger `bitsize` value that can be used together with `Scalar`, because the range for the next eligible `bitsize` (i.e., `[0, 2^256)`), contains all possible `Scalar` values.

### Changes

* Update `rust-toolchain` to `nightly-2023-01-01` and Rust edition in `Cargo.toml` to `2021`.
* Update dependencies versions: `curve25519-dalek` to `3.2.1`, `sha3` and `digest` to `0.9`, and `merlin` to `3`.
* Add `scalar_range_proof` feature, which allows `RangeProof` to be used with `Scalar`.
* Allow range proof with `bitsize` of 128 and make use of `u128` instead of `u64` in range proof.
* Extend range proof benchmarks with `u128` and `Scalar` as input.
* Extend range proof tests with `u128` and `Scalar` as input.
* Make `InnerProductProof` and `inner_product` public.

### Extras

This PR also solves issues #359 and #363.


ripa1995:
Dear @cathieyun, @hdevalence, @oleganza; 
it looks like that we are not allowed to put reviewers, so I'm adding this comment as "request for review", tagging all of you, since you are the authors of the library.

